PR #14, opened 2018-07-16
Pwielders/extend client abstraction

pwielders:


Compositor/lib/Nexus/Client/Nexus.cpp
```diff
@@ -0,0 +1,173 @@
+#include "Implementation.h"
+
+#include <EGL/eglext.h>
+
+#include <GLES2/gl2.h>
+#include <GLES2/gl2ext.h>
+
+#include <string.h>
+
+#define BACKEND_BCM_NEXUS_NXCLIENT 1
```
modeveci: Do we need hard-coded, isn't it coming from egl.pc ?

PR #16, opened 2018-07-27
Add Nexus ready systemd notification option

rjpdasilva:
When the Nexus Server is run by the Compositor Plugin and the WPE
Framework service is brought up via systemd, it is important that the
Nexus Server notifies systemd when it is ready for accepting
connections, so that systemd can wait for this notification before
starting other services that need to connect to Nexus. This is a feature
that is implemented in Broadcom's reference 'nxserver' service (when
it's compiled with the 'ENABLE_SD_NOTIFY' option).

This way, add a new 'WPEFRAMEWORK_NEXUS_SERVER_SD_NOTIFY' option to the
compositor plugin (disabled by default) which, when enabled, performs
the mentioned systemd notification, by using the systemd's library
'sd_notifyf()' function. The notification is done in the new
'Platform::NotifyReady()' method and called right after
'nxserver_ipc_init()' succeeds from 'Platform::Platform()'.

With the new option enabled, there's a new dependency to the systemd
library, so add a new cmake find module ('FindSystemd.cmake') for
making sure the dependency can be satisfied and get the required
elements (includes, libs and definitions).

Signed-off-by: Ricardo Silva <ricardo.silva@gbtembedded.com>

### DIFF
--- a/Compositor/lib/NexusServer/CMakeLists.txt
+++ b/Compositor/lib/NexusServer/CMakeLists.txt
@@ -12,12 +12,16 @@ option(WPEFRAMEWORK_PLUGIN_COMPOSITOR_BOXMODE "Allows for selecting a boxmode (N
 option(WPEFRAMEWORK_PLUGIN_COMPOSITOR_GRAPHICS_HEAP_SIZE "Change graphic heap of driver (Nexus only).")
 option(WPEFRAMEWORK_PLUGIN_COMPOSITOR_INSTALL_NEXUS_SERVER "Enable to install the server to staging." OFF)
 option(WPEFRAMEWORK_NEXUS_SERVER_HAS_EXTENDED_INIT "Enable to support unauthenticated nxclients." ON)
+option(WPEFRAMEWORK_NEXUS_SERVER_SD_NOTIFY "Enable systemd notification when nexus server is ready." OFF)
 
 message("Setup WPEFramework${TARGET} v${VERSION}...")
 
 find_package(Nexus REQUIRED)
 find_package(NXServer REQUIRED)
 find_package(Core REQUIRED)
+if(WPEFRAMEWORK_NEXUS_SERVER_SD_NOTIFY)
+find_package(Systemd REQUIRED)
+endif()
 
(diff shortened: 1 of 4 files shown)


wouterlucas:
Looks good to me! Very welcome change for systemd based devices.
Pinging @pwielders and @bramoosterhuis for final review.

pwielders:
The idea is good and we need to make sure that other systems align with the WPEFramework system. However as systemd is not part of the core and should be enabled upon user request (WPEFramework is meant to be lean and mean), suggest to move the systemd notifications (probably more might be popping up during the project) to a dedicated plugin. System architectures looking for integration with systemd should than enable the systemdplugin.
This also will only make the systemd plugin dependent on the systemd libraries and not the nexus server.
The notification that the platform is up and running is already available in the form of the platform event.

pwielders:
Happy to take the pull request into account for now but hope we can move the functionality to a plugin ASAP. Lets discuss it as well as you are on-site next week.

rjpdasilva:
Hi @pwielders, @wouterlucas , thank you for your comments.

> suggest to move the systemd notifications (probably more might be popping up during the project) to a dedicated plugin. System architectures looking for integration with systemd should than enable the systemd plugin.
This also will only make the systemd plugin dependent on the systemd libraries and not the nexus server.

Good reasoning, fully agree.

> The notification that the platform is up and running is already available in the form of the platform event.

I did noticed the `IStateChange` which I guess is what you mean that triggers the platform ready event, but didn't want to mess with that with the limited knowledge I still have about the plugins implementation.

> Happy to take the pull request into account for now but hope we can move the functionality to a plugin ASAP.

We do need the functionality now, but we can live with this patch being applied through Yocto (I'm already doing that), so feel free to drop this pull request. Main goal was to transmit our problem and how it can be solved, and I think this has been achieved.

woutermeek:
After some discussion we will be moving this functionality to this [plugin](https://github.com/WebPlatformForEmbedded/WPEFrameworkPlugins/tree/master/SystemdConnector)

PR #18, opened 2018-08-02
Anjali/ds resolution hal

anjalirajan:
New WPEFramework Plug-in for Output Resolution - RDK HAL.
Implemented the same by using a dummy ds-hal library. For the proper compilation of the plug-in, please make ensure that the following files are present in the desired location.

./output/host/usr/arm-buildroot-linux-gnueabihf/sysroot/usr/include/dsVideoPort.h
./output/host/usr/arm-buildroot-linux-gnueabihf/sysroot/usr/lib/libds-hal.so
./output/host/usr/arm-buildroot-linux-gnueabihf/sysroot/usr/include/dsError.h
./output/host/usr/arm-buildroot-linux-gnueabihf/sysroot/usr/include/dsTypes.h
./output/host/usr/arm-buildroot-linux-gnueabihf/sysroot/usr/include/dsHost.h

Following are the procedure followed by TATA team to test the DSResolution Plug-in using Dummy ds-hal library.
1.Uncomment https://github.com/WebPlatformForEmbedded/WPEFrameworkPlugins/blob/anjali/DSResolutionHAL/DSResolution/CMakeLists.txt#L43
2. make wpeframework-plugins-reconfigure
3. It will generate the dummy library, configuration script & configuration file in the following path
a. ./output/host/usr/arm-buildroot-linux-gnueabihf/sysroot/usr/lib/libds-hal.so
b. ./output/target/usr/DummyHALConfig.sh
c. ./output/target/usr/HALConfig
4. Boot up the board.
5. cd /usr
6. chmod 777 DummyHALConfig.sh
7. Change options as required, options are listed in the script itself.
8. Run the script
9. Restart the WPEFramework
10. Enable tracing if required for DSResolution Plug-in.


DSResolution/src/DSHAL.cpp
```diff
@@ -0,0 +1,191 @@
...
+};
+
+std::vector<DummyConfig> configuraton;
+
+void initConfig()
+{
+    DummyConfig dummyConfig;
+    std::string config;
+    char* token;
+    std::ifstream configFile("/usr/HALConfig");
```
ladelberg: It is generally preferred to avoid hard coding filenames in such lines. From a code maintenance perspective, one would prefer #define HAL_CONFIG_FILE "/usr/local/etc/WPEFramework/HALConfig"
or more cleanly
#define WPEFRAMEWORK_CONFIG_LOCATION "/usr/local/etc/WPEFramework"
#define WPEFRAMEWORK_HAL_CONFIG_FILE "RDK_DS_HALSettings.conf"

Also note config files should not be placed in /usr as a general practice.

DSResolution/src/DSHAL.cpp
```diff
@@ -0,0 +1,191 @@
...
+                else {
+                    dummyConfig.value = token;
+                }
+                token = strtok (NULL, "=");
+            }
+            configuraton.push_back(dummyConfig);
+            delete [] conf;
+        }
+    }
+    configFile.close();
```
ladelberg: You risk setting the ios_base:fail bit if your file was not actually present and successfully opened, which is not a big deal.  But your calling function does not seem to check the bits nor does your init actually report if it can't find/access the desired file.  You could report better errors from dsHostInit if you knew the file itself was not found/inaccessible or trapped the ios_base bits.

DSResolution/src/DSHAL.cpp
pwielders: Can we not make this part of the config. The idea behind a plugin that everything that can be configured is in the COnfig. Default value, is oke to be hard codeded, but it should be possible to override. Feed it than using in the function signature.

DSResolution/DSResolutionHAL.cpp
```diff
@@ -0,0 +1,53 @@
...
+         }
+    }
+}
+
+DSResolutionHAL::~DSResolutionHAL()
+{
+    _isOperational = false;
+}
+
+uint16_t DSResolutionHAL::GetDSResolution()
```
pwielders: Can this not be a const method?

DSResolution/DSResolutionHAL.cpp
```diff
@@ -0,0 +1,53 @@
...
+    _isOperational = false;
+}
+
+uint16_t DSResolutionHAL::GetDSResolution() 
+{
+    TRACE(Trace::Information, (_T("GetDSResolution")));
+    int _vopHandle = 0;
+    dsVideoPortResolution_t resolution;
+    if (dsGetResolution(_vopHandle,&resolution) == dsERR_NONE) 
+        return (uint16_t)resolution.pixelResolution;
```
pwielders: Prefer C++ casting, this is C++ code ;-)

DSResolution/DSResolution.h
```diff
@@ -0,0 +1,75 @@
...
+        virtual Core::ProxyType<Web::Response> Process(const Web::Request& request) override;
+
+    private:
+        Core::ProxyType<Web::Response> GetMethod(Core::TextSegmentIterator& index);
+        Core::ProxyType<Web::Response> PostMethod(Core::TextSegmentIterator& index, const Web::Request& request);
+
+    private:
+        uint8_t _skipURL;
+        PluginHost::IShell* _service;
+        Core::ProxyType<DSResolutionHAL> _controller;
```
pwielders: Should the lifetime of the DSResolutionHAL extend the plugin lifetime? If not, why not a composit?

DSResolution/DSResolution.cpp
```diff
@@ -0,0 +1,135 @@
...
+    {
+        TRACE(Trace::Information, (string(_T("GetMethod"))));
+        Core::ProxyType<Web::Response> result(PluginHost::Factories::Instance().Response());
+        result->ErrorCode = Web::STATUS_BAD_REQUEST;
+        result->Message = _T("Unsupported GET request.");
+
+        if (index.IsValid() == true) {
+            if (index.Next() && index.IsValid()) {
+                Core::ProxyType<Web::JSONBodyType<DSResolution::Config>> response(jsonBodyDataFactory.Element());
+                if (index.Remainder() == _T("GetDSResolution")) {
```
pwielders: Why call the method GetDSResolution, doet the method GET not already define this.

DSResolution/DSResolution.cpp
```diff
@@ -0,0 +1,135 @@
...
+    {
+        TRACE(Trace::Information, (string(_T("PostMethod"))));
+        Core::ProxyType<Web::Response> result(PluginHost::Factories::Instance().Response());
+        result->ErrorCode = Web::STATUS_BAD_REQUEST;
+        result->Message = _T("Unsupported POST requestservice.");
+
+        if (index.IsValid() == true) {
+            if (index.Next() && index.IsValid()) {
+                Core::ProxyType<Web::JSONBodyType<DSResolution::Config>> response(jsonBodyDataFactory.Element());
+                if (index.Remainder() == _T("SetDSResolution") && (request.HasBody())) {
```
pwielders: Could call this Resolution as well. Will make the interface symetrical, REST Get -> Gets, Post -> will set

DSResolution/DSResolution.cpp
```diff
@@ -0,0 +1,135 @@
...
+    {
+        ASSERT(service != nullptr);
+        ASSERT(_controller.IsValid() == false);
+
+        string result;
+
+        _skipURL = static_cast<uint8_t>(service->WebPrefix().length());
+        _service = service;
+
+        _controller = DSResolutionHAL::Create();
```
pwielders: No need to create this on the HEAP and make it Reference counted, looks like it 's lifetime is equal to the lifetime of the plugin.

DSResolution/src/DSHAL.cpp
```diff
@@ -0,0 +1,191 @@
...
+            }
+        }
+    }
+    return dsRet;
+}
+
+dsError_t  dsSetResolution(int handle, dsVideoPortResolution_t *resolution)
+{
+    dsError_t dsRet = dsERR_GENERAL;
+    printf("%s:%s:%d Resolution : [%d]\n",__FILE__,__func__,__LINE__,(uint16_t)resolution->pixelResolution);
```
pwielders: Suggest using TRACE or TRACE_L1. WPEFRamework determines direction of logging information. printfs are suppressed in daemon mode. In foreground, it will clutter up the output.

DSResolution/DSResolution.cpp
pwielders: Also take a look at an already existing interface (RESTfull API) in the compositor:
https://github.com/WebPlatformForEmbedded/WPEFrameworkPlugins/blob/master/Compositor/Compositor.cpp#L5-L17
and 
https://github.com/WebPlatformForEmbedded/WPEFrameworkPlugins/blob/master/Compositor/Compositor.cpp#L155-L160
and
https://github.com/WebPlatformForEmbedded/WPEFrameworkPlugins/blob/master/Compositor/Compositor.cpp#L168-L190
Please do not use an index number into an array to set the resolutions. It is there to be backwards compatible for new deplyment, we would liek to just use the resolution as a text field

DSResolution/src/DSHAL.cpp
anjalirajan: Avoided the usage of both printf() TRACE or TRACE_L1, hence it is specific to ds-hal dummy library.

anjalirajan:
Thank you.

Best Regards
Anjali Rajan

Specialist - Broadcast | Embedded Product Design
TATA ELXSI
Neyyar Techno park Campus Trivandrum 695 581 India
Tel +91 9946880547
www.tataelxsi.com<http://www.tataelxsi.com/>

PR #24, opened 2018-08-21
Cturner/add both endians

charlie-ht:
The first commit here is from the old pwielders/SessionDecoupling branch, not sure if you want this on master, it appears to be Nexus related, my change is just to add both endian swaps of the key id in playready XML parses, since we've seen both in our test cases.


albertd:
https://github.com/WebPlatformForEmbedded/WPEFrameworkPlugins/commit/f8c8861938a8298dfe1e6c36a86ccdc2ed057b36

PR #34, opened 2018-10-14
[CMAKE] WPEWebKit-2.22 packageconfig file is written as wpe-webkit-0.…

wouterlucas:
…1.pc and not wpe-webkit.pc. As such add both in the package config check module function so we should find either one of them for backward compatilibity.

Signed-off-by: wouterlucas <wouter@wouterlucas.com>


wouterlucas:
Closing, has issues with wpe-20170728, need to find more elegant solution. The calvaris branch is a good start: https://github.com/WebPlatformForEmbedded/WPEFrameworkPlugins/tree/calvaris/wpe-2.22

PR #37, opened 2018-10-29
[COMPOSITOR] Adding support to enable hdcp

johansaji:
this allows option to enable and confiure the HDCP support in the nexus layer


Compositor/lib/Nexus/NexusServer/NexusServer.cpp
```diff
@@ -132,6 +144,10 @@ namespace Broadcom {
             , SagePath()
             , SVPType(NONE)
             , Resolution(Exchange::IComposition::ScreenResolution::ScreenResolution_720p)
+            , HDCPLevel(HDCP_NONE)
+            , HDCPVersion(AUTO)
+            , HDCP1xBinFile()
```
pwielders: The selection for 1 or 2 is made I guess in the HDCPVersion as this variable does not allow support for two vlavours, I think that the BinFile will also come in 1 flavor, so lets reduce the BinFile parameters to 1 as well e.g. HDCPBinFile ?

johansaji: @pwielders if HDCPVersion is set to Auto, Nexus will first try to authenticate with the HDCP2.2 bin file, Display device doesn't support, the nexus will try to use HDCP1.X bin file authentication. In that case we need to specify both files.

PR #38, opened 2018-10-30
Introduce INotification to Resume clients

modeveci:
Putting clients in suspend mode is linked with Standby Query, but there was no resume call for clients when we awake the system from standby. "Awaking" is handled by platform at the first place, it changes nexus layer from standby to on state but there is no notification mech. to inform power plugin for resume. So we may need to notify Power plugin based on INotification. 

I also updated WpeFramework to introduce IPower::INotification https://github.com/WebPlatformForEmbedded/WPEFramework/tree/odeveci/power_resumed

If this PR is Ok, I will also merge that branch. 

Thanks
Ozgur

### DIFF
--- a/Power/Power.cpp
+++ b/Power/Power.cpp
@@ -31,7 +31,8 @@ static Core::ProxyPoolType<Web::JSONBodyType<Power::Data> > jsonResponseFactory(
 
         // Receive all plugin information on state changes.
         _service->Register(&_sink);
-        
+        _power->Register(&_sink);
+
         _power->Configure(_service->ConfigLine());
 
     } else {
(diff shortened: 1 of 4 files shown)


pwielders:
Agree with @HaseenaSainul but also agree with the commit of the INotification. I think we need the INotification to expose power changes to the outside world (between Plugin an plugin users) So do think the interface should be added (if we make it statChange and as a parameter the power state.
This will resolve the issue @hand, where the BCM imlementation needs to know when the resume should be called.
However between the implementation of BRCM and the plugin we should change it to a Callback as the relation here a a 1-1 relation. 1-1 Relations should be a callback (do not require lists), Notifications support a 1-N relation and thus are indeed more heavy but allow more clients to "observe".

PR #39, opened 2018-10-30
Anjali/monitor extension

anjalirajan:
Implementation for Monitor Plug-In extension.

The default restart threshold is 3 and is configured in the Monitor.json & if required, we can update it by using the Controller UI application.
And the report is shown in the Notification area of the Controller UI.


Monitor/Monitor.h
```diff
@@ -455,6 +473,25 @@ namespace Plugin {
...
 
             public:
+                inline void IncrRestartCount ()
+                {
+                    _restartCount++;
+                }
+                inline void ResetRestartCount ()                {
+                    _restartCount = 0;
+                }
+                inline uint32_t GetRestartCount ()
```
pwielders: Get & SET are not requirequired. This should be obvious from the signature. Please use const on a getter so was expecting the methods:
inline uint32_t RestartCount() const;
inline void RestartLimit(const uint32_t value);
inline uint32_t RestartLimit() const;

const is a usefull keyword to signify the purpose of a method or parameter. So although the const uint32_t parameter does not significally change the way is used it tells the user of this method that that parameter is an invariant and might be a temporary.

woutermeek:
Looks good, merging.

PR #40, opened 2018-11-02
Anjali/remote control key release

anjalirajan:
Implementation for Remote Control Plugin Release Key.

### DIFF
--- a/RemoteControl/LinuxDevice.cpp
+++ b/RemoteControl/LinuxDevice.cpp
@@ -116,26 +116,22 @@ namespace Plugin {
                 if (GetTimerStarted()) {
                     while (_timer.isTimeout(TIMEOUT) && !_flag) {
                         if (GetTimerStarted()) {
-                            if (_pressed) {
+                            if (_pressed)
                                 _timedOutCode = _code;
-                            }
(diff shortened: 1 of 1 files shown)


pwielders:
I think the location of this code in the whole architecture is a bit incorrect. The assumption is that the fixes you are making is to overcome a missing release in case a RELEASE key is not followed after a PRESSED key for a given time.
This issue is applicable, I guess, to all key producers, not only the linux Input device, but also, for example a web interface, that loses it connection after sending the PRESSED.
Assuming this is the reason, I would have solved it in the core (Key handling routine in the VirtualInput. 
 There the generic behaviour of generating the Repeat key event is done as well. If you do it there, you also do not need to create additional system resources, like a timer (which in turn is a thread) since you can now piggy back on the repeat key generation event. If this has been triggered for more than N times (which stands for M seconds) you signal a Key Release your self.
This allows for lower resource usage and it enables this functionality for any misbehaving key producer.

pwielders:
What is the requirement for this code extension ?

anjalirajan:
Implemented  the same in VirtualInput and raised a new PR.

PR #48, opened 2018-12-17
[CMAKE] Use CMAKE_FIND_ROOT_PATH as STAGING_DIR is BR specific and CM…

wouterlucas:
…AKE_SYSROOT is not set by Yocto.

Yocto has support for multiple sysroots (native and cross) and as such the CMAKE_SYSROOT can only hold one. Use CMAKE_FIND_ROOT_PATH, as we did before, to find the root paths in the build system. Besides CMAKE_SYSROOT should be used as a value downstream (to the compiler to set --sysroot) not by the build system to find the staging area. As per https://cmake.org/cmake/help/v3.0/variable/CMAKE_SYSROOT.html
Don't blame the buildsystem for silly cmake choices.

Signed-off-by: Wouter-lucas van Boesschoten <wouter@wouterlucas.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,10 @@
 cmake_minimum_required(VERSION 2.8)
 
-include($ENV{STAGING_DIR}${CMAKE_INSTALL_PREFIX}/include/cmake/tools.cmake)
+foreach(searchPath ${CMAKE_FIND_ROOT_PATH})
+    if(EXISTS "${searchPath}/${CMAKE_INSTALL_PREFIX}/include/cmake/tools.cmake")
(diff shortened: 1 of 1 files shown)


pwielders:
Simplicity is key to any good development. Unfortunately neither CMake nor Yocto adhere to this rule. Why can an a build system not set a sysroot either native or cross as it knows whether it will build native or cross (it is capable of selecting the proper compiler cross/native, it is not capable of selecting the proper sysroot) interesting aspect.
As one 1 line now becomes a while I find something try something, it is becoming fuzzy logic again. Besides that, this now also needs to be copied to all the stand alone Plugins, as they adhere to the same concept.
I do agree that CMake makes silly changes, buit it does not make it worse or better than Yocto makeing silly choices.
For now lets merge, I will try to find and see if the fuzzy logic can be made explicit again.

PR #59, opened 2019-01-29
WifiControl Plugin : PUT config operation hangs

arifcep:
This commit will fix bug in wificontrol plugin config put operation
The hang was cause due improper use of synchronization lock mechanism
in the code.
GetKey API internally tries to acquire the lock which is already acquired
by put method which results in hang. Put method and GetKey are in two
different thread.
Test Steps:

Put a configuration:
curl -d '{"ssid":"SSID", "psk":"***", "key":"WPA"}' -v -H "Content-Type: application/json" -v -X PUT http://ip:8082/Service/WifiControl/Config

### DIFF
--- a/WifiControl/Controller.h
+++ b/WifiControl/Controller.h
@@ -945,6 +945,8 @@ public:
         EnabledContainer::iterator entry (_enabled.find(SSID));
 
         if (entry != _enabled.end()) {
+             //Config call internally calls Lock so unlocking here to avoid dead lock
+	     _adminLock.Unlock();
             // This is an existing config. Easy Piecie.
             result = Config (Core::ProxyType<Controller>(*this), SSID);
         }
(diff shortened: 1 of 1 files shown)


WifiControl/Controller.h
```diff
@@ -945,6 +945,8 @@ class Controller : public Core::StreamType<Core::SocketDatagram> {
         EnabledContainer::iterator entry (_enabled.find(SSID));
 
         if (entry != _enabled.end()) {
+             //Config call internally calls Lock so unlocking here to avoid dead lock
+	     _adminLock.Unlock();
```
pwielders: The fact that you need to unlock is probably due to the fact that the assignment operator in the next step (result = Config()) call CopyProperties and CopyProperties starts another communication with WPASupplicant.
Same goes for the last Unlock moved to before the signment operator to assign the config.  
I think these CopyProperties in the Config object should be reduced (at least should not happen in the CopyConstructor, it should in the assignment operator) but I am most interested in the callstack that you had when this happend
Could you share these calstacks so we can vene more optimieze these calls..

PR #67, opened 2019-02-09
[WPE-221][SYSTEMCOMMANDS] Added plugin implementation.

pawel-stanek-ml:


pawel-stanek-ml:
Doing force-push was a bit scary but I found out on the Internet this is the way to do github rebases. Anyway. Rebased to the latest master. Uff.

pwielders:
Will be picked up by Marcel.

SystemCommands/SystemCommands.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+            result->ErrorCode = Web::STATUS_BAD_REQUEST;
+            result->Message = _T("Unsupported request for the [SystemCommands] service.");
+
+            if ((request.Verb == Web::Request::HTTP_PUT ||
+                request.Verb == Web::Request::HTTP_POST) &&
+                request.HasBody() && index.Next()) {
+
+                if (index.Current().Text() == "USBReset") {
+                    auto requestParams = request.Body<Web::JSONBodyType<Core::JSON::ArrayType<Param>>>();
```
MFransen69: Why not have the device name also in the URL? Multiple advantages: no need for a body, no need to parse json (why is the device an array as you only take the first one?), easier to compose for the one doing the call...

SystemCommands/SystemCommands.cpp
```diff
@@ -0,0 +1,79 @@
...
+                    Core::JSON::ArrayType<Param>::ConstIterator paramsIter = requestParams->Elements();
+                    string device;
+                    while (paramsIter.Next()) {
+                        auto& param = paramsIter.Current();
```
MFransen69: If is has to be an array, why not just an array of strings?

SystemCommands/SystemCommands.cpp
```diff
@@ -0,0 +1,79 @@
...
+                            break;
+                        }
+                    }
+
+                    if (device.empty()) {
+                        result->ErrorCode = Web::STATUS_BAD_REQUEST;
+                        result->Message = _T("No device name given to USBReset system command.");
+                    } else {
+                        uint32_t status = _implementation->USBReset(device);
+                        if (status != Core::ERROR_NONE) {
```
MFransen69: If I'm right we never send an OK result for the request always an STATUS_BAD_REQUEST (although more Plugins have this problem and the return code is ignored anyway in a lot of the js code ;) )

SystemCommands/SystemCommands.h
```diff
@@ -0,0 +1,59 @@
...
+        ~SystemCommands() override = default;
+
+        BEGIN_INTERFACE_MAP(SystemCommands)
+            INTERFACE_ENTRY(PluginHost::IPlugin)
+            INTERFACE_ENTRY(PluginHost::IWeb)
+            INTERFACE_AGGREGATE(Exchange::ISystemCommands, _implementation)
+        END_INTERFACE_MAP
+
+        //   IPlugin methods
+        const string Initialize(PluginHost::IShell* service) final;
```
MFransen69: We never use the final command so much. Why explicitly prohibit overriding this? Perhaps nice for a discussion over a beer ;) I would only use this express (and make it impossible of course) do not override, terrible things will happen if you do.

SystemCommands/ISystemCommandsImplementation.h
```diff
@@ -0,0 +1,26 @@
...
+    class ISystemCommandsImplementation : public Exchange::ISystemCommands {
+    public:
+        ISystemCommandsImplementation() = default;
+        ~ISystemCommandsImplementation() override = default;
+
+        BEGIN_INTERFACE_MAP(ISystemCommandsImplementation)
+            INTERFACE_ENTRY(Exchange::ISystemCommands)
+        END_INTERFACE_MAP
+
+        uint32_t USBReset(const std::string& device) final;
```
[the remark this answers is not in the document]
pawel-stanek-ml: This should be the only one USBReset thingy IMO so matter of opinion and also didin't see a use case not to stop here.

SystemCommands/SystemCommands.cpp
pawel-stanek-ml: Stings won't neccessary work for any type of a command

SystemCommands/SystemCommands.cpp
pawel-stanek-ml: Prepared to be ready for other commands also one with multiple params so I take at some point putting everything in URL won't cut it, right?

MFransen69: Eh, but every command can just have it's unique URL and params, similar to the C++ Interface or am I missing something? So you could have .../SystemCommand/USBReset/[devicename]  and .../SystemCommand/FormatHDD/[drive] (not sure if we want this one ;) ) and if we would have a very complex command .../SystemCommand/VeryComplexNeedsLotsOfDifficultParams that one could then have a json body.

SystemCommands/SystemCommands.cpp
MFransen69: See above, I assumed it was specific for the USBReset command...

SystemCommands/ISystemCommandsImplementation.h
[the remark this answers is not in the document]
MFransen69: Agree, as I mentioned it is perhaps also a matter of how you look at it. My opinion up to now has always been why would I closeup the code for extension because I don't see the use case  at the moment (although I agree that I do not see any reason why you would extend here ;) ). Personally I only use final to indicate that you are really not expected to extend as things will go wrong if you do, not to indicate that it is just not expected.

SystemCommands/SystemCommands.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+        SERVICE_REGISTRATION(SystemCommands, 1, 0);
+        static Core::ProxyPoolType<Web::JSONBodyType<Core::JSON::ArrayType<SystemCommands::Param>>> jsonBodyResponseFactory(2);
+
+        const string SystemCommands::Initialize(PluginHost::IShell* service)
+        {
+            string result;
+            _skipURL = static_cast<uint8_t>(service->WebPrefix().length());
+            uint32_t pid = 0;
+            _implementation = service->Root<Exchange::ISystemCommands>(pid, 2000, _T("ISystemCommandsImplementation"));
```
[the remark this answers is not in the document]
pwielders: The interface returned her: service->Root<Exchange::ISystemCommands>(pid, 2000, _T("ISystemCommandsImplementation")); Is reference counted so it means if you do not "Release this interface, whether it is in process or out of process, it will leak if you do not destruct (Release) it in the Deinitialize. Also take care, see othere examples, to be Robust in case the Deinitialize does not report that the object is destructed, force a destrcut, force kill the process...

SystemCommands/SystemCommands.cpp
```diff
@@ -0,0 +1,73 @@
...
+            if (!_implementation)
+                result = "Couln't create ISystemCommands instance.";
+            return result;
+        }
+
+        void SystemCommands::Deinitialize(PluginHost::IShell* service)
+        {
+            if (_implementation != nullptr) {
```
MFransen69: Hi Pawel, had a quick look and I notice that cleanup is not completely correct for an OOP plugin, there could be an external process running so that also needs to be taken into account. Please have a look at for example the WebServer Deinitialize code (WebServer.cpp). Here you will see that WebServer::Deinitialize will take into account that if the Release will not lead to a destruction succeed it will kill the process if needed. For that reason it also has a void WebServer::Deactivated(RPC::IRemoteProcess* process) implementation, in case the process dies (e.g. crashes for example). I think you want a similar implementation here as well.

SystemCommands/SystemCommands.cpp
```diff
@@ -0,0 +1,73 @@
...
+            index.Next();
+
+            result->ErrorCode = Web::STATUS_BAD_REQUEST;
+            result->Message = _T("Unsupported request for the [SystemCommands] service.");
+
+            if ((request.Verb == Web::Request::HTTP_PUT ||
+                request.Verb == Web::Request::HTTP_POST) && index.Next()) {
+
+                if (index.Current().Text() == "USBReset") {
+                    string device = index.Next() ? index.Current().Text() : "";
```
MFransen69: I guess nitpicking ;) (no need to change this) but I would have just used an if( index.Next() ). Easier to understand what is going on, you only need one if and not unnecessarily contruct the empty device. But the empty case is the exception so  not that important.

pawel-stanek-ml: Saw than checking the status and force killing. It's needed only becaue because I call Root() method so it can quietly change to be OOP some day?

pawel-stanek-ml: Ah. This was meant to be in the other comment.

MFransen69: > Saw than checking the status and force killing. It's needed only becaue because I call Root() method so it can quietly change to be OOP some day?

Indeed, of course it is just a configuration option. Although the default is just false it would be completely  allowed for somebody to just edit the json and make it true (OOP). And he would expect it to work as expected. (that is why I mentioned it is okay with me if we allow it to be OOP but it is not strictly necessary. Easier to implement if you do not offer that possibility ;)

pawel-stanek-ml: Yup. I give up on a new fancy way and do it old school

SystemCommands/SystemCommands.config
```diff
@@ -0,0 +1,2 @@
+set (autostart true)
+set (preconditions Platform)
```
MFransen69: Is there a reason for the precondition to be Platform? In the past we typically used this to reset the network interface as the network stack could not be setup correctly, so that would be early in the startup. Platform is very late in the startup cycle.

pawel-stanek-ml: My thinking was: Platfrom is something generic saying we're up and running and  since this is to do System (so platform related) stuff itr should depend on it. Moreover most of the plugin depend on it.

pawel-stanek-ml: Streamer, SystemCommector, IOConnector, RemoteControl, TVControl, DeviceInfo, GsgccCliennt, Power, WifiControl, DHCPServer. All depend on Platform

PR #71, opened 2019-02-19
[WiFiControl] fix for pure virtual termination

HaseenaSainul:


HaseenaSainul:
> Why is this mandatory? I think it has no use to wait for the delay. Could you try to test without this line ?

Otherwise, during the deactivation, seeing a zombie of [wpa_supplicant]

PR #73, opened 2019-03-06
[WPE-213][Monitor] Improve restarting logic.

pawel-stanek-ml:
The previous restart limit used for both memory and oprerational
failures were split into two: separate limit for memory and
operational failures. Moreover a restarting window (seconds)
setting was added. The new logica causes that if a plugin
hits the restart limit within the window it's restarted no more.


pawel-stanek-ml:
The review was done during a skype call with @pwielders. All comments have been addressed this is why this was merged to master (this has also been agreed on during the call)

PR #86, opened 2019-03-25
[TRACE CONTROL] Add JSONRPC API

Damian-Danylko:


TraceControl/TraceControl.h
```diff
@@ -6,7 +6,12 @@ namespace WPEFramework {
 
 namespace Plugin {
 
-    class TraceControl : public PluginHost::IPlugin, public PluginHost::IWeb {
+namespace {
+    constexpr auto* kStatusMethodName = _T("status");
```
pwielders: Can we for these very simple types just use the actual type, TCHAR in this case. Also as we do not expect this value to be a nullptr, the most verbose definition would be:
constexpr TCHAR kStatusMethodName[] = _T("status");

TraceControl/TraceControl.h
```diff
@@ -810,6 +815,25 @@ namespace Plugin {
         };
         class Data : public Core::JSON::Container {
         public:
+            class StatusParam final : public Core::JSON::Container {
+            public:
+                StatusParam()
+                    : Core::JSON::Container()
+                    , Module("")
```
pwielders: If it is empty string, just use the default constructor ("") is not needed.

TraceControl/TraceControl.h
Damian-Danylko: Done

TraceControl/TraceControl.h
Damian-Danylko: Done

PR #91, opened 2019-03-29
fix linking wpe-0.2

dwrobel:


pwielders:
@bramoosterhuis oke to merge ?

bramoosterhuis:
I guess this is a workaround for a faulty package config file? I think we need a proper find module for libwpe is needed instead.

dwrobel:
> BTW, why is libwpe is not found by using a find module, like glib and wpewebkit?

Why do you think it was not found. Please have a look at the linker error above it just says that it couldn't found `-lwpe-0.2`. That means that the `pkg-config` `Libs:` were correctly used to pass `-lwpe-0.2` but no one passed `-L${libdir}` to the linker.

For completeness here is the content of my `wpe-0.2.pc` installed in `/home/dwrobel/projects/webkit/WebPlatformForEmbedded-2.22/1/_install/usr/lib/pkgconfig/wpe-0.2.pc`:

```
prefix=/home/dwrobel/projects/webkit/WebPlatformForEmbedded-2.22/1/_install/usr
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${exec_prefix}/lib

Name: wpe-0.2
Description: The wpe library
Version: 1.0.0
Requires: xkbcommon
Cflags: -I${includedir}/wpe-0.2
Libs: -L${libdir} -lwpe-0.2
```

bramoosterhuis:
Yeah, so in your development environment, the binaries are installed on an exotic location. Its extra code for only for a specific development environment.  I think it better to instruct your cmake also to look in your specific location instead of solving it here. See the buildroots toolchainfile.cmake or you can use buildroot to build your stuff for your native platform without even polluting your dev machine.

bramoosterhuis:
and about the find module see:

https://github.com/WebPlatformForEmbedded/WPEFrameworkPlugins/pull/100#pullrequestreview-229943691

dwrobel:
>Yeah, so in your development environment, the binaries are installed on an exotic location.

It's not exotic as it something you can officially pass using [DESTDIR](https://cmake.org/cmake/help/v3.14/envvar/DESTDIR.html) during the `install`.

> I think it better to instruct your cmake...

`Cmake` already has the instruction how to link properly using `libwpe` from its `pkg-config` file as it contains `-L${libdir}`. It just need to use it and IMHO recommending `buildroot` is not the recommended solution :-) - it's probably one of the possibly workaround.

pwielders:
@bramoosterhuis @dwrobel  ready to merge now or will it break the build ?

bramoosterhuis:
I guess it will not break the build, @dwrobel do you still need this?

dwrobel:
> @dwrobel do you still need this?

At the time I last checked it, we need it.

> I guess it will not break the build,

It looks that CI flow would be also really handy, any plans for establishing it?

pwielders:
@dwrobel  there is CI but only on the master branch and some specific customer branches. It used to be Jenkins and currently we are using buildbot.
But if I know on forehand that it does not build, I will not merge it, but the expectation is that it builds (according to Bram) I am taking a chance ;-)

PR #92, opened 2019-04-02
Damiand/rpcjson/locationsync/1

Damian-Danylko:


LocationSync/LocationSync.cpp
```diff
@@ -96,8 +96,8 @@ namespace Plugin {
             result->ContentType = Web::MIMETypes::MIME_JSON;
             result->Body(Core::proxy_cast<Web::IBody>(response));
         } else if (request.Verb == Web::Request::HTTP_POST) {
-
             if (index.Next()) {
+                index.Next();
```
pwielders: If the index.Next does not result in a valid position, the index.Current can not be called, so make this part of the if on the next line. In debug the Current in line 101 will assert if the Next is pointing to an invalid situation.

Damian-Danylko: Done. It was my intention but I put Next in wrong line. Now shoudl be fine.

PR #106, opened 2019-04-26
[OpenCDMi] allign configuration options for playready read-dir and st…

artur-gebicz-ml:
…ore-location


OpenCDMi/CMakeLists.txt
```diff
@@ -4,8 +4,8 @@ set(MODULE_NAME ${NAMESPACE}${PLUGIN_NAME})
 find_package(ocdm REQUIRED)
 find_package(${NAMESPACE}Plugins REQUIRED)
 
-set(PLUGIN_OCDM_PLAYREADY_READ_DIR "/root/Netflix/dpi/playready" CACHE STRING "Playready read-dir")
-set(PLUGIN_OCDM_PLAYREADY_STORE_LOCATION "/root/Netflix/dpi/playready/storage/drmstore" CACHE STRING "Playready store-location")
+set(PLUGIN_OPENCDMI_PLAYREADY_READ_DIR "" CACHE STRING "Playready read-dir")
+set(PLUGIN_OPENCDMI_PLAYREADY_STORE_LOCATION "" CACHE STRING "Playready store-location")
```
MFransen69: Have you tested this on the Raspberry Pi? The default is now missing and that is used by all RPi configuration if I'm right.

artur-gebicz-ml: @MFransen69  Yes, you are right. This would provide regression until configuration option is provided from buildroot (please see PR for nf51 branch https://github.com/WebPlatformForEmbedded/buildroot/pull/106/commits/afb38a93087499eb2a846d0502a3b545966eed17). 

I have reverted default value for those two flags, so now it will not prevent regression on other platforms and will enable possibility to set it from BR environment in future like in nf51 case.

PR #108, opened 2019-04-29
[wificontrol] Add jsonrpc handling.

pawel-stanek-ml:


pawel-stanek-ml:
I think I see some issues. Closing for now.

PR #113, opened 2019-05-09
[WPE-276][Wificontrol] Fixes for various issues.

pawel-stanek-ml:
- Scanning not ending
- Status not updating properly
- Not being able to delete first config entry
- Invalid data being stored in a config

### DIFF
--- a/WifiControl/Controller.cpp
+++ b/WifiControl/Controller.cpp
@@ -200,7 +200,7 @@ namespace WPASupplicant {
                 if ((event == CTRL_EVENT_CONNECTED) || (event == CTRL_EVENT_DISCONNECTED) || (event == WPS_AP_AVAILABLE)) {
                     _statusRequest.Event(event.Value());
                     Submit(&_statusRequest);
-                } else if ((_networks.size() == 0) && (event.Value() == CTRL_EVENT_SCAN_RESULTS) && (_scanRequest.Set() == true)) {
+                } else if ((event.Value() == CTRL_EVENT_SCAN_RESULTS) && (_scanRequest.Set() == true)) {
                     Submit(&_scanRequest);
                 } else if ((event == CTRL_EVENT_BSS_ADDED) || (event == CTRL_EVENT_BSS_REMOVED)) {
 
--- a/WifiControl/Controller.h
+++ b/WifiControl/Controller.h
@@ -267,6 +267,9 @@ namespace WPASupplicant {
 #endif // __DEBUG__
             virtual void Completed(const string& response, const bool abort) = 0;
 
+        protected:
+            void Reset() { _request.clear(); }
+
         private:
             string _request;
 #ifdef __DEBUG__
@@ -294,6 +297,7 @@ namespace WPASupplicant {
             inline bool Activated()
             {
                 if (_scanning == false) {
+                    Reset();
                     _scanning = true;
                     return (true);
                 }
@@ -417,6 +421,14 @@ namespace WPASupplicant {
             inline void Event(const events value)
             {
                 _eventReporting = value;
+            }
+
+            void Reset()
+            {
+                _bssid = 0;
+                _ssid.clear();
+                _pair = 0;
+                _key = 0;
             }
 
         private:
@@ -1068,7 +1080,7 @@ namespace WPASupplicant {
             // See if the given SSID is enabled.
             EnabledContainer::iterator entry(_enabled.find(SSID));
 
-            if ((entry != _enabled.end()) && (entry->second.Id() != 0)) {
+            if ((entry != _enabled.end()) && (entry->second.Id() != static_cast<uint32_t>(~0))) {
 
                 // This is an existing config. Easy Piecie.
                 //Config internally calls Lock so unlocking here to avoid dead lock
@@ -1226,10 +1238,14 @@ namespace WPASupplicant {
                 if ((exchange.Wait(MaxConnectionTime) == false) || (exchange.Response() != _T("OK"))) {
 
                     result = Core::ERROR_ASYNC_ABORTED;
+                } else {
+                    _statusRequest.Reset();
                 }
 
                 Revoke(&exchange);
             } else {
+                if (_statusRequest.SSID() == SSID)
+                    _statusRequest.Reset();
                 _adminLock.Unlock();
             }
 
@@ -1466,7 +1482,7 @@ namespace WPASupplicant {
 
             Submit(&exchange);
 
-            if ((exchange.Wait(MaxConnectionTime) == false) && (exchange.Response() != _T("FAIL"))) {
+            if ((exchange.Wait(MaxConnectionTime) == false) || (exchange.Response() == _T("FAIL"))) {
                 result = Core::ERROR_ASYNC_ABORTED;
             } else {
                 result = Core::ERROR_NONE;


pawel-stanek-ml:
To clarify there're 2 Reset()s. One is on the StatusRequest and the other one is on the ScanRequest.
StatusRequest::Reset() makes sure the status is proper when the network gets disconnected.
The other makes sure ScanRequest::Set() has a chance to succeed after call to Scan(). This is needed for the scan operation to finish - because Submit() for that and for scan operation it's protected by if checking if Set() succeeded. Having that said, Reset() serves a different purpose.
However there is the code which changes the behavior you mentioned. Namely it's _networks.clear() in L#817. I could remove that. My only concern then in the case the available networks have changed since the last scan (i.e. some disappeared in a meanwhile) won't a dead networks be incorrectly listed?   
 
> First of all, good fixes 
> 
> Jus one general question on the Reset:
> The idea is that if I start a new scan, the old information is maintained by default. So the SSID's found during the last scan should still be in a list.
> Do agree that you want to have a the possibility to do a "clean" scan but that has to be explicit.
> So what I am asking, is the functionality still such that a new scan does not "clean" the found SSID, but it can be achieved by first explicitly calling Clear (or Reset) and then start a scan.
> How you want to scan (first clearing or not) is domain logic and will be coded in javascript and so you should be able to do this over the JSON RPC interface.

pawel-stanek-ml:
If BSS_REMOVED event covers the case of disappearing network L#817 could be removed and I don;t think a clean scan is really needed.

> To clarify there're 2 Reset()s. One is on the StatusRequest and the other one is on the ScanRequest.
> StatusRequest::Reset() makes sure the status is proper when the network gets disconnected.
> The other makes sure ScanRequest::Set() has a chance to succeed after call to Scan(). This is needed for the scan operation to finish - because Submit() is needed for that and for scan operation it's protected by the if statement checking if Set() succeeded. Having that said, Reset() serves a different purpose.
> However there is the code which changes the behavior you mentioned. Namely it's _networks.clear() in L#817. I could remove that (or transform to a separate JSONRPI API). My only concern then is if - in the case the available networks have changed since the last scan (i.e. some disappeared in a meanwhile) dead networks be incorrectly listed?
> 
> > First of all, good fixes
> > Jus one general question on the Reset:
> > The idea is that if I start a new scan, the old information is maintained by default. So the SSID's found during the last scan should still be in a list.
> > Do agree that you want to have a the possibility to do a "clean" scan but that has to be explicit.
> > So what I am asking, is the functionality still such that a new scan does not "clean" the found SSID, but it can be achieved by first explicitly calling Clear (or Reset) and then start a scan.
> > How you want to scan (first clearing or not) is domain logic and will be coded in javascript and so you should be able to do this over the JSON RPC interface.

pwielders:
Off course in a happy and ideal world all is fine and all works fine, the BSS_ADDED and BSS_REMOVED will keep the list up-to date. However, and I hope this does not come as a big surprise :-), sometimes software crashes (e.g. WPASupplicant). I would like to design a system that can cope with these none-happy day scenarios.
By definition the domain control lies in the JavaScript world, so I do not want to put implicit behaviour in the plugin, it should slave to the requests from JSONRPC.
So suppose the following scenario takes place:
Wifi started and collected a list of networks:
- A/B/C
- wpa_supplicant crashes (May need to add support to report this to JSONRPC)
- plugin restarts wpa_supplicant
- Up to domain logic to take proper action. If a scan would start now wit a cleared list of networks or not.
  Suppose network B has a very low beacon transmit rate, might be interesting not to clear the list.
  If you want to create an absolute correct list of what the driver knows right now, you must clear the list.

Any of the last two points are valid and I do not want to enforce a decision on any of the two in the implementation, so lets make it explicit and leave it up to the user of this plugin.

pwielders:
Your concern: 
My only concern then is if - in the case the available networks have changed since the last scan (i.e. some disappeared in a meanwhile) dead networks be incorrectly listed?
Is a valid concern, but having a redundant list (that is what the network list in the WifiController is) can always be out-of-sync. However worst case that can happen is that a use connects to a "none" existing SSID, which in turn leads to an error like: "Could not connect to SSID: X".
The rationale here is that the inconsistent list should not be reolved by the plugin but by a "higher" layer. The higher layer should be able to go left or right and should try to recover from an inconsistent list (if possible) which as litle aspossible user impact.

pwielders:
Suggest to make the clear intelligent. Only clear elements that have not yet been reported by the wpa_supplicant in the scan or successive ADD/REMOVE notifications.
This way the higher layers, can, after a crash, keep the network list as it. higher layer triggers the scan, the scan completed result than allows the higher layer to say clear and on this clear, you only clear out the networks that have not announced themself yet, since the crash.

pawel-stanek-ml:
Ok. I can do it explicitly.

Now let's talk about clear() you'd like it to be:
Question #1: Who has information the supplicant has crashed (as a notification. We do not want to poll Core::Process) (for sure it's needed to pass this to JS if its the one in the control and it's not in place already)? This is must have to known the point the list gets invalidated. Also I do not think the driver is ever restarted except for WifiContol Init()/Deinit().
Question #2: Let's look at clear() usage:
- JS calls it at any point and there was no crash and all works as it should so it does nothing (which doesn't seem expected for someone who was given the control) and the list get broken on JS side (bugs there also happens and ppl do not always ready the documentation carefully and make assumptions on historical data).  
- JS knows how it works (documentation read) and knows the driver crashed so knows that the results from scan() might be at least partially invalid. So calls it before calling scan() and clears everything. This is fine. But what if JS given the new powers and knowledge comes to the conclusion: ok let's scan() and then clear() so only the networks which didn't refresh their validity are cleared (this is the whole point of the complicated implementation, no). Fine. Unless in the meanwhile between the scan ended notification (which triggers clear() call) the driver cashes again. I can imagine that in multithreaded environment it's possible C++ knows about the crash already at point it gets clear()  and since this is the call after scan() the list will be cleared completely and in this case it's not fine at all.

pawel-stanek-ml:
I updated the PR with mentioned L#817 removed so no network list clearing. The new clear() entry is to be worked out and done. I think it should be done separately since this servers the purposes it's was filed for: fixes challenges known in WifiControl  up to the point if was filled in and does not behavior of scan and network list. I could get back to this subject once I'm done with the new I started or in a meanwhile but I don't want to block this.

pwielders:
I think there is misunderstanding I probably did not explain it properly. Lets do a Skype call on monday as well.
What I expect:
1) scan ran, list as is. all well. All network entries flagged as "reported". in case a BS_ADDED/BS_REMOVE comes in, all is handled, as these events are coming from wpa_supplicant, ADDED, adds an entry (tagged as reported right away), REMOVE removes an network entry.
So far no issue list should be in sync.
2) Crash detected (wpa_supplicant connection fails, socket closed event or process gone. don't know if this is implemented, but assume it is) we also forward the fact that wpa_supplicant is down to JS via an event.
All Network entries sofar available will clear the "reported" flag, which means, Network entry might be there but we have it not yet validated by wpa_supplicnt, WifiControl will fire up wpa_supplicant again and start a scan.
3) due to the restart and the scan, networks detected come in again, each entry reported by wpa_supplicant is moved to the "reported" state again.
4) JS might now decide to say "ClearStaleNetworks". This means that Networks that have not been detected sofar (after the crash) will be removed as the are stil in the un-reported state,

Now whenever there is a crash and a new scan is started, the list is not automagically cleared and the "old" list is still available, so a UI can still be populated with networs although the moste recent scan is not completed.
I know this is not yet so fancy implemented, but if you would "clear" the nework list at the moment a new scan is started, the UI would be "empty" for a given time, that is what I want to prevent and that was my remark for which I was looking confirmation. In the implementation, if a new scan is started, do you not clear the network list.

For now the UI could force a clear of this network list by starting and stopping the plugin. Not ideal but works. The scenario drawn up above will happen in the near future and we will get a ticket for it. So we can prevent this ticket by implementing it now, or raise a ticket and implement it when it is reported :-)

pawel-stanek-ml:
I imagined it exactly as you described it and we're not there yet (no wpa_supplicant crash detection and notification, no wpa_supplicant restart on crash, no rescan on crash initiated [I'm not even sure it should since the JS would get the info about the crash and has scan() entry to do whatever it decides]) unless I can't find it in the code. This is completely separate discussion unrelated to WPE-276 I worked on which has just popped up, though. We can always talk about improvements/refactors. They are good. They are great. They are more than welcome. My question is should WPE-276 be blocked/dropped because of this? Does the scope of WPE-276 changes, then? Whatis its priority?

Scan/Rescan won't clear anything in this code. This could only happen due to L#817 because it was about clearing _networks list which is the list UI retrieves via networks() entry (and it's not cleared).

pawel-stanek-ml:
Also when thinking about the improved solution wouldn't it still be better to instead of exposing clearInvalid() API always validate the list after the crash and let the JS know about changes via an event. This would make the internal plugin info to alway be valid and up to date yet JS still may decide to retrieve or not the retrieve the updated list?

pawel-stanek-ml:
I allow myself to summarize this a bit during this sleepless - for me at least - night.
Whenever I write "API" I mean JSONRPC endpoint.The main API I'll focus on here are: scan(), networks(), networkchange and scanresults events.

The current code put to the review fixes the issues mentioned in the message commit and *doesn't* clear the networks list (Controller.h:817 and an old Controller.cpp:203 *mustn't* be in the code). Meaning that even after scan() and before any networkchange/scanresults event arrives networks() will return an old list (which could be not quite valid). If all works fine the list should update and networkchange/scanresults events will inform about it. This means no behavior change introduced and having fixes in and as that I believe this is good to go.
Detecting wpa_supplicant's crash, notifying about it, restarting the driver and eventually the latest operation it was performing before the crash (if any) and what else to do in this case is a different discussion for me. Having this in would definitely be a robustness improvement.

*** SIDE NOTE WITH A PERSONAL OPINION START ***
In order to clear the list one would have to restore L#817 in Controller.h and revert the change in if in L#203 in Controller.cpp. I think such would be better then a dead list because it's the UI which calls scan() and networks() and gets the events and even if the list is empty after scan() (and no events come due to e.g. crash) the UI can have the logic to replace whatever is currently shown with what networks() returns or not to do this (until scanresults comes) and use the cached value. If it's the crash which made the list empty it's IMO better to make it empty to indicate an error since trying to connect to something on a fake list will fail anyway until the driver is restarted and operational. All in all the plugin IMO should do whatever it can to make the list as valid as possible and UI should not care about invalidating it or not. It should only care about information endpoints/events returns.  
*** SIDE NOTE WITH A PERSONAL OPINION END ***

PR #114, opened 2019-05-13
Compositor client removed

HaseenaSainul:
Composi


Compositor/lib/Wayland/Compositor/Wayland.h
```diff
@@ -1,7 +1,7 @@
 #ifndef PROJECT_WAYLAND_H
 #define PROJECT_WAYLAND_H
 
-#include "../Client/Implementation.h"
+#include <compositorclient/Implementation.h>
```
pwielders: This implementation.h should be part of this directory and not the compositorclient. I hope it can be moved into this directory..

Compositor/lib/Wayland/Compositor/Wayland.h
```diff
@@ -1,8 +1,9 @@
 #ifndef PROJECT_WAYLAND_H
 #define PROJECT_WAYLAND_H
 
-#include <compositorclient/Implementation.h>
 #include "Module.h"
+#include <compositor/Client.h>
+#include <compositorclient/Implementation.h>
```
pwielders: compositor/Client.h is fine but the compositorclient/Implementation.h contains implementation and should be possible (I hope) to be only used locally, in this compositor server component.

Compositor/lib/Wayland/Compositor/Wayland.h
HaseenaSainul: Implementation.h is required for compositorclient/Wayland, since this is the part of client implementation. So cannot move this to compositor server

Compositor/lib/Wayland/Compositor/Wayland.h
HaseenaSainul: compositorclient/implementation.h are used as the part of  Wayland/Compositor Client.
If move it to Wayland/Compositor, there will be two copy for the same

PR #116, opened 2019-05-14
Constify OnKeyMessage() f_pszUrl parameter

dwrobel:
Helps to fix errors on OCDM-Playready like;
     error: ISO C++ forbids converting a string constant to ‘char*’
             m_piCallback->OnKeyMessage((const uint8_t *) "", 0, "");


dwrobel:
Depends on: https://github.com/WebPlatformForEmbedded/WPEFramework/pull/119

pwielders:
Agree on the change and potentially we need more fixes but we are deviating from the Frauenhofer released interface more and more.

PR #124, opened 2019-06-10
[WPE-295][wificontrol] Asure Submit() assert won't trigger.

pawel-stanek-ml:
Everytime a request is to be submitted it's checked it's not in progress
by doing Set() on it. If Set() doesn't succeed it means a request is in
progress so there's no point in scheduling a new one. Submit() is even
asserting on that. However Set() success or failure was dependent on internal
request's message being empty or not. The problem was SendData() is clearing
that meassage yet the fact the message was sent doesn't mean it's done
processing. Receiving the response is such indicator. This is why it's now
assured that until response/cancellation request's Set() will fail preventint
its re-submition.


HaseenaSainul:
@pawel-stanek-ml, this looks good. I think this is the right one to avoid sending the duplicates.

PR #125, opened 2019-06-11
[wificontrol] Fix locking

pawel-stanek-ml:


WifiControl/Controller.h
```diff
@@ -266,18 +266,20 @@ namespace WPASupplicant {
                 return (_original);
             }
 #endif // __DEBUG__
-            void Completed(const string& response, const bool abort) {
-                _settable = true;
-                OnCompleted(response, abort);
+
+            void SetProcessing(bool processing)
```
pwielders: Setters and Getters can be distinguished by their signature. For a getter the method is const, for the setter it is not const. If the signature already reveals the purpos (get or Set) why add this allso to the method name? In WPEFramework we would expect here:  "void Processing(const bool processing)"

WifiControl/Controller.h
```diff
@@ -266,18 +266,20 @@ namespace WPASupplicant {
...
+            virtual void Completed(const string& response, const bool abort) = 0;
+
         private:
-            virtual void OnCompleted(const string& response, const bool abort) = 0;
             string _request;
 #ifdef __DEBUG__
             string _original;
 #endif // __DEBUG__
-            bool _settable = true;
+            bool _settable { true };
```
pwielders: Current methods are initializing values of member variables at the constructor, which happens here as well. Why is this one initialized here ? Why not alos in the constructor list ?

PR #126, opened 2019-06-17
[json-rpc] Modified files according to new jsonrpc headers.

aswaunni:


sebaszm:
Looks good.

pwielders:
I understand from @sebaszm that this change should go together with the automatic generation of the generated header files. so we need to align this pull with those changes. Suggest @aswaunni that you merge this pull when all the other CMake changes are succefully tested as well. So leave it up to you to merge this.

aswaunni:
> I understand from @sebaszm that this change should go together with the automatic generation of the generated header files. so we need to align this pull with those changes. Suggest @aswaunni that you merge this pull when all the other CMake changes are succefully tested as well. So leave it up to you to merge this.

ok

PR #131, opened 2019-06-28
Sebaszm/combined streamer

sebaszm:
Adds posibility to compile and run both QAM and AAMP streamers simultaneously, selectable in runtime via stream type parameter of the create method. 

AAMP player slightly adjusted to utilize the AAMP events and adjusted JSON-RPC interface a bit.


HaseenaSainul:
@sebaszm Seems fine for me :)

Streamer/Administrator.cpp
```diff
@@ -0,0 +1,237 @@
...
+                            streamer.second->Name().c_str(), streamer.second->Type()));
+                } else {
+                    TRACE(Trace::Information, (_T("Initialized streamer '%s' (type %i, %i frontend(s) available)"),
+                            streamer.second->Name().c_str(), streamer.second->Type(), streamer.second->Frontends()));
+                }
+            }
+
+            _decoders = config.Decoders.Value();
+            if (_decoders > 0) {
+                _decoder_slots = new bool[_decoders];
```
pwielders: Suppose you would use a uint32_t to indicate this bool, using a bit per decoder in use or not. We could go up to 32 decoders before we run out of space... Guess, looking at the working area 32 is unlikely, so why not safe the hassle of the array and just use a uint32_t, and bits in there ?

Streamer/Administrator.cpp
```diff
@@ -0,0 +1,237 @@
...
+                ASSERT(_decoder_slots != nullptr);
+
+                if (_decoder_slots != nullptr) {
+                    memset(_decoder_slots, 0, (sizeof(*_decoder_slots)*_decoders));
```
pwielders: this is just _slots = 0; :-)

Streamer/Administrator.h
```diff
@@ -0,0 +1,85 @@
...
+            Exchange::IStream* Acquire(const Exchange::IStream::streamtype streamType);
+            void Relinquish(IPlayerPlatform* player);
+
+            uint8_t Allocate();
+            void Deallocate(uint8_t index);
+
+        private:
+            Core::CriticalSection _adminLock;
+            std::map<Exchange::IStream::streamtype, IPlayerPlatformFactory*> _streamers;
+            uint8_t _decoders;
```
pwielders: If the number of decoders are limited to 256 (which is already redicously high ;-) ) We could use 4 uint64_t for the bool to go for the allocated stuff (would not suggest so... but just to address the consern  that we might have more than 32 decoders...)

Streamer/Administrator.h
```diff
@@ -0,0 +1,85 @@
...
+        };
+
+        template<class PLAYER>
+        class PlayerPlatformRegistrationType {
+        public:
+            PlayerPlatformRegistrationType() = delete;
+            PlayerPlatformRegistrationType(const PlayerPlatformRegistrationType<PLAYER>&) = delete;
+            PlayerPlatformRegistrationType& operator=(const PlayerPlatformRegistrationType<PLAYER>&) = delete;
+
+            PlayerPlatformRegistrationType(const string& name, Exchange::IStream::streamtype streamType,
```
pwielders: Minor remarks, but just curious, why not use the ClassName (PLAYER) as the name, so you enforce that the classname is equeal to the name used in the factory, plus, if you Announce with (name, FactoryInterface), you do not need to maintain the factory, you coud than revoke the name again.

Streamer/Implementation/Aamp/PlayerImplementation.cpp
```diff
@@ -337,17 +369,29 @@ namespace Player {
...
             }
 
             _adminLock.Unlock();
             return result;
        }
+
+    } // namespace Aamp
+
+    namespace {
+        static PlayerPlatformRegistrationType<Aamp::PlayerPlatform> Register(PLAYER_NAME, Exchange::IStream::streamtype::Unicast,
```
pwielders: Nice way to solve the configuration.

Streamer/Implementation/Aamp/PlayerImplementation.h
```diff
@@ -18,12 +18,14 @@ namespace WPEFramework {
 
 namespace Player {
 
-    namespace Implementation {
+namespace Implementation {
```
pwielders: The implementation is not needed outside the scope of the source (compilation unit) lets not expose it in the header file, but move it into the PlayerImplementation.cpp. It makes sure that noone can start "abusing" the implementation. Furtheremore, we could than name (in any namescope) the Player implementation (PLAYER_NAME Aamp) and than use the ClassNameOnly(TYPE) without passing the name as a string from a preprocessor define (preprocessors are namespace less and prefer to avoid them as much as possible)

Streamer/Implementation/QAM/PlayerImplementation.h
```diff
@@ -13,9 +13,11 @@ namespace WPEFramework {
 
 namespace Player {
 
-    namespace Implementation {
+namespace Implementation {
```
pwielders: See remark on the AAMP, also here I think we could get ride of the header file, hiding all the specific impleemntation details into the compilation unit (*.cpp file)

Streamer/PlayerPlatform.h
```diff
@@ -0,0 +1,237 @@
...
+
+        struct ICallback {
+            virtual ~ICallback() {}
+
+            virtual void TimeUpdate(uint64_t position) = 0;
+            virtual void DRM(uint32_t state) = 0;
+            virtual void StateChange(Exchange::IStream::state newState) = 0;
+        };
+
+        class IPlayerPlatform
```
pwielders: Can we keep pure abstract classes (interfaces) of the struct type. Just for consistency.

Streamer/PlayerPlatform.h
```diff
@@ -0,0 +1,237 @@
...
+            virtual void Window(const Rectangle& rectangle) = 0;
+
+            virtual uint32_t Order() const = 0;
+            virtual void Order(const uint32_t order) = 0;
+
+            virtual uint32_t AttachDecoder(const uint8_t index) = 0;
+            virtual uint32_t DetachDecoder(const uint8_t index) = 0;
+        };
+
+        class IPlayerPlatformFactory
```
pwielders: See remark above

Streamer/PlayerPlatform.h
```diff
@@ -0,0 +1,237 @@
...
+            virtual void Destroy(IPlayerPlatform* player) = 0;
+
+            virtual const string& Name() const = 0;
+            virtual Exchange::IStream::streamtype Type() const = 0;
+            virtual uint8_t Frontends() const = 0;
+            virtual const string& Configuration() const = 0;
+        };
+
+        template<class PLAYER>
+        class PlayerPlatformFactoryType : public IPlayerPlatformFactory
```
pwielders: I guess, this time the slots are for the frontends :-) I guess, now we might consider a uint64 for the number of frontends, and in case the number of frontends is 0, it is unlimited (like for aamp).

Streamer/PlayerPlatform.h
```diff
@@ -0,0 +1,237 @@
...
+                            }
+
+                            Core::JSON::DecUInt8 Frontends;
+                        } config;
+
+                        config.FromString(_configuration);
+                        _max_slots = config.Frontends.Value();
+
+                        if (_max_slots > 0) {
+                            _slots = new bool[_max_slots];
```
pwielders: As for the decoders, I think the frontends will never exceed the 16 (in the near future, highest number I heard sofar on high-end boxes is 8). So I guess we could do here with a uint32_t as well. Maybe we can make this a "generic" template class, that initially is always limited to 32 bits, and if needed in the future can increase to any value? I see the same behaviour for the decoders as well, so generic, is then already used twice. Suggest to start with a simple uint32...

Streamer/PlayerPlatform.h
```diff
@@ -0,0 +1,237 @@
...
+
+                if ((index < _max_slots) && (_slots[index] == true)) {
+                    _slots[index] = false;
+                    delete player;
+                }
+            }
+
+            const string& Name() const override
+            {
+                return (_name);
```
pwielders: This could than be Core::ClassNameOnly(typeid(PLAYER).name); If we force the name of the class to be equal to the player instance.

Streamer/StreamerJsonRpc.cpp
```diff
@@ -7,6 +7,8 @@ namespace WPEFramework {
 
 namespace Plugin {
 
+    const int MAX_STREAMS = 255;
```
pwielders: Do we really want to go so high? How about starting at 32 ?

sebaszm:
Thanks @pwielders, adjusted code as per review comments/suggestions.

pwielders:
Looks very good. Impressed by the BitArray template. I suggest to move this template to the Core, might be usefull to more situations ! For me good to merge now !!!

PR #136, opened 2019-07-16
Updating TestUtility to changes in framework

mipo57:
Adapted TestUtility to work with IRemoteConnection rather than old IRemoteProcess, fixed problem with TestUtility not shutting down after deactivating plugin.


Damian-Danylko:
I will merge it, but please create JIRA ticket for Crash functionality. As I understand correctly it is not working at the moment.

PR #139, opened 2019-07-17
Ddanylko/memory monitoring/1

Damian-Danylko:


WebKitBrowser/WebKitBrowser.cpp
```diff
@@ -49,9 +49,14 @@ namespace Plugin {
 
             } else {
                 _browser->Register(&_notification);
-                _memory = WPEFramework::WebKitBrowser::MemoryObserver(_service->RemoteConnection(_connectionId));
+
+                const RPC::IRemoteConnection *connection = _service->RemoteConnection(_connectionId);
+
+                ASSERT(connection != nullptr);
```
pwielders: Prepare for receiving a nullptr. If there is something wrong, the connection might close before you pick it up (plugin crashes before you reach this point) and thus the _connectionId returns a nullptr.

Damian-Danylko: Done

Damian-Danylko: Done

PR #140, opened 2019-07-17
DHCP Client now remembers previous IP

mipo57:
1. Fixed DHCP client sending ACK instead of REQUEST
2. Added IP retention - it is no longer changing on every reboot

### DIFF
--- a/NetworkControl/DHCPClientImplementation.cpp
+++ b/NetworkControl/DHCPClientImplementation.cpp
@@ -97,16 +97,7 @@ namespace Plugin {
     {
         if (!_persistentStorage.empty()) {
             Core::File file(_persistentStorage + lastIPFileName);
-            bool isFileOpen = false;
-
-            if (file.Exists()) {
-                isFileOpen = file.Open(false);
-                file.SetSize(0);
-            } else {
-                isFileOpen = file.Create();
-            }
-
-            if (isFileOpen) {
+            if (file.Create()) {
                 IPStorage storage;
                 storage.ip_adress = last_ip.c_str();
                 storage.ToFile(file);
--- a/NetworkControl/DHCPClientImplementation.cpp
+++ b/NetworkControl/DHCPClientImplementation.cpp
@@ -30,9 +30,12 @@ namespace Plugin {
         , _offers()
         , _persistentStorage(persistentStoragePath)
     {
-        _adminLock.Lock();
          _last = Core::NodeId(ReadLastIP().c_str(), DefaultDHCPClientPort, Core::NodeId::TYPE_IPV4);
-        _adminLock.Unlock();
+        if (!_persistentStorage.empty()) {
+            if (!Core::Directory(_persistentStorage.c_str()).CreatePath()) {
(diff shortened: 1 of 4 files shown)


NetworkControl/DHCPClientImplementation.cpp
```diff
@@ -18,16 +18,21 @@ namespace Plugin {
...
         , _interfaceName(interfaceName)
         , _state(IDLE)
         , _xid(0)
         , _preferred()
+        , _last()
         , _callback(callback)
         , _offers()
+        , _persistentStorage(persistentStoragePath)
     {
+        _adminLock.Lock();
```
pwielders: During construction, I tjhink the risk of multiple concurency on this object is low. No need for locking here

NetworkControl/DHCPClientImplementation.cpp
```diff
@@ -63,5 +68,55 @@ namespace Plugin {
...
+
+        return result;
+    }
+
+    void DHCPClientImplementation::SaveLastIP(const string& last_ip) 
+    {
+        if (!_persistentStorage.empty()) {
+            Core::Directory directory(_persistentStorage.c_str());
+
+            if (directory.CreatePath()) {
```
pwielders: As this is a one time action (creating the directory) if needed lets do this in the constructor of this object. Now there is a continous check on eachSaveLastIp()

NetworkControl/DHCPClientImplementation.cpp
```diff
@@ -63,5 +68,55 @@ namespace Plugin {
...
+                Core::File file(_persistentStorage + lastIPFileName);
+                printf("SAVE FILE PATHNAME: %s", file.PathName().c_str());
```
pwielders: remove printf's the cluttter up the output with unwnated messages (in Release as well)

NetworkControl/DHCPClientImplementation.cpp
```diff
@@ -63,5 +68,55 @@ namespace Plugin {
...
+                if (file.Exists()) {
+                    file.Open(true);
+                    file.SetSize(0);
+                } else {
+                    file.Create();
```
pwielders: I think with Create you can also create a file overwrite..

NetworkControl/DHCPClientImplementation.h
```diff
@@ -140,6 +143,18 @@ namespace Plugin {
         static constexpr uint16_t DefaultDHCPServerPort = 67;
         static constexpr uint16_t DefaultDHCPClientPort = 68;
 
+        class IPStorage : public Core::JSON::Container {
+        public:
+            IPStorage()
```
pwielders: Please specify the CopyConstructor and Assignment operator as deleted.

NetworkControl/DHCPClientImplementation.cpp
mipo57: I checked and I think I would need to add a O_TRUNC flag to open() function inside Create, but I can only change access flags.

pwielders:
Good to go for me !

PR #143, opened 2019-07-23
[Compositor] Added JSON-RPC api to Compositor

mipo57:
Added JSON-RPC API to Compositor (based on old REST API)
Also, fixed a bug where ZOrder returned empty strings at the beginning

### DIFF
--- a/Compositor/CompositorJsonRpc.cpp
+++ b/Compositor/CompositorJsonRpc.cpp
@@ -1,0 +1,277 @@
+
+#include "Module.h"
+#include "Compositor.h"
+#include <interfaces/json/JsonData_Compositor.h>
+
+namespace WPEFramework {
+
+namespace Plugin {
+
+    using namespace JsonData::Compositor;
+
+    // Registration
+    //
+
+    void Compositor::RegisterAll()
+    {
+        Register<TotopParamsInfo,void>(_T("totop"), &Compositor::endpoint_totop, this);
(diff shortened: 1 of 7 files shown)


Compositor/CompositorJsonRpc.cpp
```diff
@@ -0,0 +1,277 @@
...
+    {
+        uint32_t result = Core::ERROR_NONE;
+        const string& below = params.Below.Value();
+        const string& above = params.Above.Value();
+
+        result = PutBelow(above, below);
+
+        if (result != Core::ERROR_FIRST_RESOURCE_NOT_FOUND && result != Core::ERROR_NONE) {
+            TRACE_L1("Failed to put surface of client %s\n below %s\n", below.c_str(), above.c_str());
+            result = Core::ERROR_GENERAL;
```
MFransen69: Why do we change the error code here to ERROR_GENERAL if the code is not ERROR_FIRST_RESOURCE_NOT_FOUND? If somebody adds an error to the PutBelow function later on he will not be aware that this piece of code is here. Is only ERROR_GENERAL allowed as return value (but then I don't understand why ERROR_FIRST_RESOURCE_NOT_FOUND is allowed)

(same rematk applies multiple times in this file)

sebaszm: I believe this is to prevent forwarding of an undocumented error to the JSON-RPC client. We forsee that FIRST_RESOURCE_NOT_FOUND can happen, anything else lands as GENERAL.

PR #148, opened 2019-07-24
fix PSSH Playready XML parser

goruklu:
We have seen some content (Blacklist, Episode 15 on Amazon Prime) which has a PSSH box that was much bigger than the XML string. It was not being parsed properly


goruklu:
Init segment containing the problematic PSSH data
[init.mp4.zip](https://github.com/WebPlatformForEmbedded/ThunderNanoServices/files/3429292/init.mp4.zip)

PR #149, opened 2019-07-25
[DropbearServerPlugin] Initial version for review

s-rapolu:
DropbearServer plugin implementation related changes in WPEFrameworkPlugins (ThunderNanoServices) repository

### DIFF
--- a/DropbearServer/DropbearServer.cpp
+++ b/DropbearServer/DropbearServer.cpp
@@ -1,0 +1,96 @@
+#include "DropbearServer.h"
+
+namespace WPEFramework {
+namespace Plugin {
+
+    SERVICE_REGISTRATION(DropbearServer, 1, 0);
+
+    const string DropbearServer::Initialize(PluginHost::IShell* service)
+    {
+        _skipURL = static_cast<uint8_t>(service->WebPrefix().length());
+        return string();
+    }
+
+    void DropbearServer::Deinitialize(PluginHost::IShell* service)
+    {
+    }
+
+    string DropbearServer::Information() const
+    {
+        return string();
+    }
+
+    void DropbearServer::Inbound(Web::Request& request)
+    {
+    }
+
+    Core::ProxyType<Web::Response> DropbearServer::Process(const Web::Request& request)
+    {
+        ASSERT(_skipURL <= request.Path.length());
+
+        Core::ProxyType<Web::Response> result(PluginHost::Factories::Instance().Response());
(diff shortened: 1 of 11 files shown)


DropbearServer/DropbearServer.cpp
```diff
@@ -17,42 +17,76 @@ namespace Plugin {
...
+    // GET ../Window <- Return the Rectangle in which the player is running.
+    // POST ../Create/<Type> <- Create an instance of a stream of type <Type>, Body return the stream index for reference in the other calls.
+    // PUT ../<Number>/Load <- Load the URL given in the body onto this stream
+    // PUT ../<Number>/Attach <- Attach a decoder to the primer of stream <Number>
+    // PUT ../<Number>/Speed/<speed> <- Set the stream as a speed of <speed>
+    // PUT ../<Number>/Detach <- Detach a decoder to the primer of stream <Number>
+    // PUT ../<Number>/Window/X/Y/Width/Height <- Set Window size
+    // DELETE ../<Number> <- Drop the stream adn thus the decoder
+    //
+    // POST/PUT	<- StartService
```
pwielders: Do not expose starting and stoppoing of a service through the RESTFull interface. This is already part of the Thunder framework. Activating and Deactivating a plugin.

DropbearServer/DropbearServer.cpp
```diff
@@ -17,42 +17,76 @@ namespace Plugin {
...
+    // POST/PUT	<- StopService
```
pwielders: See remark above. This is Deactivating the plugin.

pwielders:
Implementation looks good, with the exception of the RESTFul/JSONRPC starting and stopping of the service. This really needs to be removed.
Furthermore, I think it might be interesting Food for Thought if this should not be called the SecureShellServer plugin with multiple (for now only 1) implementations of a SSH server implementation. Here is the DropBear implementation but what about a tinyssh.org iimplementation ?
Should fit behind a the suggested interface and might even make the system smaller and less resource consuming. Even think that the server could be run in-process, to even more reduce the resource usage, but that is just a configuration option :-)

DropbearServer/DropbearServer.cpp
s-rapolu: Agreed. I'll make the necessary corrections.

DropbearServer/DropbearServer.cpp
s-rapolu: Agreed

PR #153, opened 2019-07-30
[DHCPServer] Fix lease time handling

sebaszm:
Attempts to fix lease time handling.

### DIFF
--- a/DHCPServer/DHCPServerImplementation.h
+++ b/DHCPServer/DHCPServerImplementation.h
@@ -492,7 +492,6 @@ namespace Plugin {
             }
             inline void Base(const std::string& serverName, const uint32_t server, const CoreMessage& message, const uint32_t router, const uint32_t dns)
             {
-
                 // Server message handling RFC 2131 section 4.3
                 _replyAddress = message.giaddr.s_addr;
                 _ciaddr = message.ciaddr.s_addr;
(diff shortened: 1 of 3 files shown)


DHCPServer/DHCPServerImplementation.h
```diff
@@ -752,24 +750,46 @@ namespace Plugin {
                     result = Create(scratchPad.Id(), scratchPad.RequestedIP());
```
artur-gebicz-ml: If the client will request the address from the outside of pool range it will be created anyway and will also be granted to the client, which should not happen. Such address should be rejected by the server and the server should offer address from withing the range

DHCPServer/DHCPServerImplementation.h
```diff
@@ -752,24 +750,46 @@ namespace Plugin {
...
+                        _lastIssued = ip;
+                        break;
+                    }
+                }
 
-                if (poolSize != 0) {
-                    result = Create(scratchPad.Id(), _lastIssued);
+                if (result == nullptr) {
+                    // Still not found a free slot, attempt picking up one of the expired ones
+                    for (ip = _minAddress; ip <= _maxAddress ; ip++) {
```
artur-gebicz-ml: Looks like the _lastIssued is initialized with the minimal available address. In such situation the first available address will not be acquired ever. It will always start with next one. Please correct that loop

DHCPServer/DHCPServerImplementation.h
artur-gebicz-ml: If the client will request the address from the outside of pool range it will be created anyway and will also be granted to the client, which should not happen. Such address should be rejected by the server and the server should offer address from withing the range

PR #157, opened 2019-08-14
[Compositor] NexusServer: enable reading IR input mode as string from…

artur-gebicz-ml:
… config file

### DIFF
--- a/Compositor/lib/Nexus/NexusServer/NexusServer.cpp
+++ b/Compositor/lib/Nexus/NexusServer/NexusServer.cpp
@@ -33,38 +33,42 @@ namespace Broadcom {
         { Exchange::IComposition::ScreenResolution::ScreenResolution_2160p60Hz, NEXUS_VideoFormat_e3840x2160p60hz }
     };
 
-    static const std::map<std::string, NEXUS_IrInputMode> irInputModeLookup = {
-        { "NEXUS_IrInputMode_eTwirpKbd", NEXUS_IrInputMode_eTwirpKbd },
-        { "NEXUS_IrInputMode_eSejin38KhzKbd", NEXUS_IrInputMode_eSejin38KhzKbd },
-        { "NEXUS_IrInputMode_eSejin56KhzKbd", NEXUS_IrInputMode_eSejin56KhzKbd },
-        { "NEXUS_IrInputMode_eRemoteA", NEXUS_IrInputMode_eRemoteA },
-        { "NEXUS_IrInputMode_eRemoteB", NEXUS_IrInputMode_eRemoteB },
-        { "NEXUS_IrInputMode_eCirGI", NEXUS_IrInputMode_eCirGI },
-        { "NEXUS_IrInputMode_eCirSaE2050", NEXUS_IrInputMode_eCirSaE2050 },
-        { "NEXUS_IrInputMode_eCirTwirp", NEXUS_IrInputMode_eCirTwirp },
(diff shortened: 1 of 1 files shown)


Compositor/lib/Nexus/NexusServer/NexusServer.cpp
```diff
@@ -33,6 +33,43 @@ namespace Broadcom {
         { Exchange::IComposition::ScreenResolution::ScreenResolution_2160p60Hz, NEXUS_VideoFormat_e3840x2160p60hz }
     };
 
+#define BEGIN_ENUMERATOR_STRING_MAP(ENUM_TYPE, NAME) static const std::map<std::string, ENUM_TYPE> NAME = {
+#define ENTRY_ENUMERATOR_STRING_MAP(ENUM) { #ENUM, ENUM },
```
pwielders: Please use the version @sebaszm created for the "Error" lookup. If required, nmove these templates (not a preprocessor shizzle) to a header file so everyone can use it.

Compositor/lib/Nexus/NexusServer/NexusServer.cpp
```diff
@@ -177,7 +214,7 @@ namespace Broadcom {
         }
 
     public:
-        Core::JSON::DecUInt16 IRMode;
+        Core::JSON::String IRMode;
```
pwielders: Sonce it is an enum, use the JSONEnum<TYPEDEF> So you can also use the Enum values here.

Compositor/lib/Nexus/NexusServer/NexusServer.cpp
pwielders: There is also the possibility of using the Core::EnumerateType (also used by the JSON variant). So internally you can still work with the enum value and the JSON will automagically convert the text enum verion to the "real" enum value..

pwielders: Could be that you need to "combine" it with the Sebas solution where the enum is automagically tyranslated uin to text. See @sebaszm created for the "Error" lookup. If required, nmove these templates (not a preprocessor shizzle) to a header file so everyone can use it.

pawel-stanek-ml: I think that using JSON EnumType as you suggested below will do the trick since:
- the value comes from the config anyway
- the value should be a string in the config and converted to an enum in the code. This is exactly what EnumType is for.

Compositor/lib/Nexus/NexusServer/NexusServer.cpp
artur-gebicz-ml: Good point, Will change it

Compositor/lib/Nexus/NexusServer/NexusServer.cpp
artur-gebicz-ml: @pwielders usign @sebaszm  solution for error lookup is the other way around, so could not be used directly. I think that JSON EnumType  is the way to go

Compositor/lib/Nexus/NexusServer/NexusServer.cpp
artur-gebicz-ml: Changed.

Compositor/lib/Nexus/NexusServer/NexusServer.cpp
artur-gebicz-ml: Changed.

PR #160, opened 2019-08-20
[WifiControl] BSSID is uint64_t and it's written to a JSON string.

pawel-stanek-ml:
Via implicit conversions with String(bool quoted) involved.
The mentioned Ctor will be made explicit as a separate change.


mipo57:
Good catch :)

PR #171, opened 2019-08-30
[WPE-342] Fix assertion due to rescheduling status request.

pawel-stanek-ml:


WifiControl/Controller.cpp
```diff
@@ -198,8 +198,14 @@ namespace WPASupplicant {
                 TRACE(Communication, (_T("Dispatch message: [%s]"), message.c_str()));
 
                 if ((event == CTRL_EVENT_CONNECTED) || (event == CTRL_EVENT_DISCONNECTED) || (event == WPS_AP_AVAILABLE)) {
+                     _adminLock.Lock();
                     _statusRequest.Event(event.Value());
```
pwielders: I guess this _statusRequest.Event() action should also, only happen if it is not in progress. Otherwise we might be creating a race condition here.

pawel-stanek-ml: Right. Fixed.

pawel-stanek-ml:
In case this is to be merged  please SQUASH the fixup commit when merging.

WifiControl/Controller.h
```diff
@@ -757,8 +759,9 @@ namespace WPASupplicant {
                     }
                     else if (SetKey("autoscan", "periodic:120") != Core::ERROR_NONE) {
                         _error = Core::ERROR_GENERAL;
-                    }
-                    else {
+                    } else {
+                        const bool set = _statusRequest.Set();
+                        ASSERT(set == true || !"StatusRequest::Set failed yet the request has just been constructed. Is must be settable.");
```
MFransen69: Ah, you got me there ;) Thought why not just call Set in the assert as it is looks like a getter to get the Set state. But indeed it has a purpose here ;)

pawel-stanek-ml:
Definitely. Sounds good, Marcel. Thx.

PR #175, opened 2019-09-05
Decouple the Injected Bundle from the WebKit plugin

pwielders:


WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -220,6 +217,24 @@ namespace Plugin {
...
+        std::unique_ptr<char[]> buffer(new char[bufferSize]);
+        size_t stringLength = WKStringGetUTF8CString(wkStringRef, buffer.get(), bufferSize);
+        return Core::ToString(buffer.get(), stringLength - 1);
+    }
+    static std::vector<string> ConvertWKArrayToStringVector(WKArrayRef array) {
+            size_t arraySize = WKArrayGetSize(array);
+
+            std::vector<string> stringVector;
+
+            for (unsigned int index = 0; index < arraySize; ++index) {
```
MFransen69: Small thing but if arraySize could be significant a strngVecor.Reserver(arraySize) would make it more efficient (less realocations)

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -220,6 +217,24 @@ namespace Plugin {
...
+                stringVector.push_back(WKStringToString(static_cast<WKStringRef>(WKArrayGetItemAtIndex(array, index))));
```
MFransen69: Expect WKStringToString to return a std::string? In that case use an emplace_back?

WebKitBrowser/InjectedBundle/Tags.cpp
```diff
@@ -0,0 +1,11 @@
+#include "Tags.h"
+
+namespace WPEFramework {
+
+namespace Tags {
+
+const char* Config = "Config.";
```
MFransen69: Guess you don't want the pointer to be changed as well? Could consider const char* const to express that

PR #176, opened 2019-09-05
[JSON RPC] Add JSON documentation for Power

Damian-Danylko:


Power/Power.json
```diff
@@ -0,0 +1,96 @@
+{
```
MFransen69: As discussed let's not put this file here.

Damian-Danylko: Done for all plugins

PR #181, opened 2019-09-09
Fix to avoid WebKit restart after enabling Monitor plugin

HaseenaSainul:


WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -1426,6 +1426,10 @@ namespace WebKitBrowser {
                 // In the end we check if all bits are 0, what means all mandatory processes are still running.
                 requiredProcesses = (0xFFFFFFFF >> (32 - requiredChildren));
 
+                if (_children.Count() < 2) {
```
pawel-stanek-ml: It'd be better to use requiredChildren instead of hardcoded values. I saw there's more places like this, though.

HaseenaSainul: Thanks @pawel-stanek-ml , magic number updated with const expression : https://github.com/WebPlatformForEmbedded/ThunderNanoServices/pull/181/commits/51286343f1125463b342ec7a7983768ecf07ce96

wouterlucas:
Validated patch on Yocto with 72604a0 board, works great!

PR #185, opened 2019-09-21
[WPE-357][WPE-356][DHCPClient] Full implementation of minimal dhcp protocol

mipo57:
[WPE-357]
The socket is now properly closed when either DHCP successfully assing IP or a maximum number of retries were exceeded.

[WPE-356]
DHCPImplementation is now not aware of permanent storage. Its sole responsibility is to implement DHCP requests and process responses.  Moreover, now the only DHCPRequest is sent if IP was saved, which is correct behaviour in such a case. Full DHCPDisover is sent only if DHCPRequest fails.  Also, multiple interfaces are supported now.

[DHCPClient] - Full implementation
Previously IP was assigned only after receiving information, that this IP is available. Result of the request was ignored. Now IP is assigned only after it is successfully given by DHCP server. If the server refuses to assign IP, other offers are tried or another DHCPDiscover is issued if no other options are available.

### DIFF
--- a/NetworkControl/DHCPClientImplementation.cpp
+++ b/NetworkControl/DHCPClientImplementation.cpp
@@ -18,24 +18,29 @@ namespace Plugin {
         return (Core::NodeId(sockaddr_broadcast));
     }
 
-    DHCPClientImplementation::DHCPClientImplementation(const string& interfaceName, ICallback* callback, const string& persistentStoragePath)
+    DHCPClientImplementation::DHCPClientImplementation(const string& interfaceName, DiscoverCallback discoverCallback, RequestCallback claimCallback)
         : Core::SocketDatagram(false, Core::NodeId(_T("0.0.0.0"), DefaultDHCPClientPort, Core::NodeId::TYPE_IPV4), RemoteAddress(), 1024, 2048)
         , _adminLock()
         , _interfaceName(interfaceName)
         , _state(IDLE)
+        , _serverIdentifier(0)
         , _xid(0)
         , _preferred()
-        , _last()
-        , _callback(callback)
-        , _offers()
-        , _persistentStorage(persistentStoragePath)
+        , _discoverCallback(discoverCallback)
+        , _claimCallback(claimCallback)
+        , _unleasedOffers()
+        , _leasedOffers()
     {
-         _last = Core::NodeId(ReadLastIP().c_str(), DefaultDHCPClientPort, Core::NodeId::TYPE_IPV4);
-        if (!_persistentStorage.empty()) {
-            if (!Core::Directory(_persistentStorage.c_str()).CreatePath()) {
-                TRACE_L1("Could not create a NetworkControl persistent storage directory");
-            }
+        Core::AdapterIterator adapters;
+        while ((adapters.Next() == true) && (adapters.Name() != _interfaceName)) /* INTENTIONALLY LEFT EMPTY */
+                    ;
+
+        if (adapters.IsValid() == true) {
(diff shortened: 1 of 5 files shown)


NetworkControl/DHCPClientImplementation.cpp
```diff
@@ -18,24 +18,29 @@ namespace Plugin {
...
+        , _claimCallback(claimCallback)
+        , _unleasedOffers()
+        , _leasedOffers()
     {
-         _last = Core::NodeId(ReadLastIP().c_str(), DefaultDHCPClientPort, Core::NodeId::TYPE_IPV4);
-        if (!_persistentStorage.empty()) {
-            if (!Core::Directory(_persistentStorage.c_str()).CreatePath()) {
-                TRACE_L1("Could not create a NetworkControl persistent storage directory");
-            }
+        Core::AdapterIterator adapters;
```
pwielders: Why not use the constructor with the name? Than the iterator is automatically pointing to the right interface, and there is no code duplication.

NetworkControl/DHCPClientImplementation.h
```diff
@@ -297,12 +421,64 @@ namespace Plugin {
...
+                    
+                    _gateway = options.gateway;
+                    _broadcast = options.broadcast;
+                    _dns = options.dns;
+
+                    // Get as much info from options as possible...
+                    if (options.netmask.IsSet()) 
+                        _netmask = options.netmask.Value();
+                    else 
+                        TRACE_L1("DHCP message without netmask information");
```
pwielders: Lets make these TRACE_L1 -> TRACE statements so we can also turn them on if we have a release build.

NetworkControl/DHCPClientImplementation.h
```diff
@@ -493,25 +678,88 @@ namespace Plugin {
             return (result);
         }
 
-        inline Iterator Offers() const
+        inline void AddOffer(const Offer& offer, bool leased) {
```
pwielders: Although technically there is no added value in making the leased parameter const, agree, it is an indication to the user of the method that this method is not a changing variable so it can be made const, so I prefer to be as accurate on the interface as possible, so suggest to also make the leased parameters in the next calls const.

NetworkControl/DHCPClientImplementation.h
```diff
@@ -370,13 +551,16 @@ namespace Plugin {
...
             uint32_t _renewalTime; /* renewal time in seconds */
             uint32_t _rebindingTime; /* rebinding time in seconds */
+            uint32_t _id; /* unique offer identifier */
         };
 
-        typedef Core::IteratorType<const std::list<Offer>, const Offer&, std::list<Offer>::const_iterator> Iterator;
-        typedef Core::IDispatchType<const string&> ICallback;
+        typedef Core::IteratorType<std::list<Offer>, Offer&, std::list<Offer>::iterator> Iterator;
+        typedef Core::IteratorType<const std::list<Offer>, const Offer&, std::list<Offer>::const_iterator> ConstIterator;
+        typedef std::function<void(Offer&)> DiscoverCallback;
```
pwielders: Can the Offer parameter not be const ?

NetworkControl/DHCPClientImplementation.h
```diff
@@ -370,13 +551,16 @@ namespace Plugin {
...
+        typedef std::function<void(Offer&, bool)> RequestCallback;
```
pwielders: Same request ;-) Can't offer be const?

PR #186, opened 2019-09-23
Sebaszm/streamer face update

sebaszm:
implement the new IStream APIs, some tidying and memory leak fixes

### DIFF
--- a/Streamer/Element.h
+++ b/Streamer/Element.h
@@ -1,0 +1,113 @@
+#pragma once
+
+#include "Module.h"
+
+#include <interfaces/IStream.h>
+
+namespace WPEFramework {
+
+namespace Player {
+
+    namespace Implementation {
+
+        struct ElementInfo {
+            ElementInfo(Exchange::IStream::IElement::type type) 
+                : Type(type) 
+            {
+            }
+
+            Exchange::IStream::IElement::type Type;
+        };
+
+        class Element : public Exchange::IStream::IElement {
+        public:
+            Element() = delete;
+            Element(const Element&) = delete;
+            Element& operator=(const Element&) = delete;   
+
+            Element(const ElementInfo& info)
+                : _elementInfo(info)
+            {                    
+            }
+            ~Element() override
+            {
+            }
+            Exchange::IStream::IElement::type Type() const override
+            {
+                return _elementInfo.Type;
+            }
+
+            BEGIN_INTERFACE_MAP(Element)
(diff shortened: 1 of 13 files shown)


Streamer/Frontend.h
```diff
@@ -66,41 +66,41 @@ namespace Player {
...
+                uint32_t Release() const override
                 {
                     if (Core::InterlockedDecrement(_referenceCount) == 0) {
                         delete this;
                         return (Core::ERROR_DESTRUCTION_SUCCEEDED);
                     }
                     return (Core::ERROR_NONE);
                 }
-                virtual uint8_t Index() const
+                uint8_t Index() const
```
pwielders: No override ?

Streamer/Frontend.h
```diff
@@ -233,31 +233,39 @@ namespace Player {
                 }
                 return (result);
             }
-            virtual uint8_t Index() const
+            uint8_t Index() const
```
pwielders: Maybe here the override ?

Streamer/Frontend.h
```diff
@@ -318,6 +318,18 @@ namespace Player {
...
             }
+            uint32_t LastError() const override
+            {
+                _adminLock.Lock();
+                ASSERT(_player != nullptr);
+                uint32_t result = _player->LastError();
+                _adminLock.Unlock();
+                return (result);
+            }
+            IStream::IElement::IIterator* Elements()
```
pwielders: override?

Streamer/Element.h
```diff
@@ -0,0 +1,113 @@
...
+
+#include <interfaces/IStream.h>
+
+namespace WPEFramework {
+
+namespace Player {
+
+    namespace Implementation {
+
+        struct ElementInfo {
```
pwielders: Why not use the namescoping by nestin. Is this struct not a part of the Element class? So I guess it could be nested in the Element class.

Streamer/Frontend.h
sebaszm: definitely

Streamer/Frontend.h
sebaszm: No, this method is not part of the interface.

Streamer/Element.h
sebaszm: Player implementations do not create IElement. This is similar to IGeometry/Rectangle concept.

PR #188, opened 2019-09-27
Srapolu/secureshell

MFransen69:


s-rapolu:
Here are the features implemented:

1. Get SSH client session information from native Dropbear through REST API  and JSON-RPC
2. Close individual session or all active sessions
3. Implemented IClient interface.
4. Variable configuration parameters are supported.
4. Integrated plugin with native Dropbear APIs.

Please review the latest implementation of SSH plugin and let us know your feedback.

PR #191, opened 2019-10-01
Development/wpe 249

Damian-Danylko:


SysLogRedirect/Module.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#ifndef MODULE_NAME
+#define MODULE_NAME Plugin_SystemCommands
```
pawel-stanek-ml: Wring name.

SysLogRedirect/SysLogRedirect.h
```diff
@@ -0,0 +1,360 @@
...
+
+#include <sys/inotify.h>
+#include <errno.h>
+#include <iostream>
+#include <fstream>
+
+namespace WPEFramework {
+namespace Plugin {
+
+    class FileSystemObserver : public Core::IResource {
```
pawel-stanek-ml: This being a singleton is wrong. In case more files are to be monitored. plugin instances will share this which look wrong.

SysLogRedirect/SysLogRedirect.h
```diff
@@ -0,0 +1,360 @@
...
+                    length = read(_iNotifyFd, _eventBuffer, sizeof(_eventBuffer));
+
+                    if ((length != static_cast<uint16_t>(~0)) && (length > 0)) {
+                        const struct inotify_event* event = reinterpret_cast<const struct inotify_event *>(_eventBuffer);
+                        int wd = event->wd;
+
+                        _adminLock.Lock();
+
+                        FileObserverMap::iterator index (_observedEntries.begin());
+                        while ( (index != _observedEntries.end()) && (index->second.wd != wd) ) { index++; }
```
pawel-stanek-ml: find_if

SysLogRedirect/SysLogRedirect.h
```diff
@@ -0,0 +1,360 @@
...
+
+                        if (index != _observedEntries.end()) {
+                            std::ifstream file (index->first);
+
+                                if (file) {
```
pawel-stanek-ml: I have dejavu here

SysLogRedirect/SysLogRedirect.h
```diff
@@ -0,0 +1,360 @@
...
+        Core::CriticalSection _adminLock;
+        int _iNotifyFd;
+        FileObserverMap _observedEntries;
+        bool _fullFileMode;
+        uint8_t _eventBuffer[(sizeof(struct inotify_event) + NAME_MAX + 1)] __attribute__ ((aligned(8)));
+    };
+
+    class SysLogRedirect : public PluginHost::IPlugin {
+    private:
+        class TextChannel : public Core::StreamTextType<Core::SocketDatagram, Core::TerminatorCarriageReturn> {
```
pawel-stanek-ml: This class looks useless. Use Core::StreamTextType instead

SysLogRedirect/SysLogRedirect.h
```diff
@@ -0,0 +1,360 @@
...
+            virtual void StateChange()
+            {}
+
+            void Emmit(const string& text)
+            {
+                Submit(text);
+            }
+        };
+
+        class OnChangeSink : public FileSystemObserver::INotification {
```
pawel-stanek-ml: Not needed. Make SysLogRedired to derives from IFileObserver::INotificaton directly. I'd remove IFileObserver::INotificaton class too BTW and replace it by callback.

SysLogRedirect/Module.h
Damian-Danylko: Done

SysLogRedirect/SysLogRedirect.h
Damian-Danylko: Unfortunatly this object has 3 pure virtual methods so I have to leave it as it is:
        virtual void Received(string& text) = 0;
        virtual void Send(const string& text) = 0;
        virtual void StateChange() = 0;

SysLogRedirect/SysLogRedirect.h
Damian-Danylko: Done

SysLogRedirect/SysLogRedirect.h
Damian-Danylko: if we are monitoring one file per plugin map is not needed I will remove it.

SysLogRedirect/SysLogRedirect.h
pawel-stanek-ml: Ouch. It seems to this class if perfectly usable without them so I'd go for changing it there so they're not abstract but it's not withing scope here. Closing

SysLogRedirect/SysLogRedirect.cpp
```diff
@@ -11,24 +11,17 @@ namespace Plugin {
         config.FromString(service->ConfigLine());
 
         Core::NodeId logNode(config.Destination.Binding.Value().c_str(), config.Destination.Port.Value());
-        _logOutput = new TextChannel(logNode);
-        _logOutput->Open(0);
+        _logOutput.reset(new TextChannel(logNode));
```
pwielders: How about no pointer at all! The classes your are using have been designed to become have a composited relation. Also this requires a new, lets only new if the plugin is created.

pawel-stanek-ml: That would be ideal indeed but TechChannel requires Core::NodeId to be given to the CTRO (base class requires it) and creation of that needs config.

pwielders: 1) TextChannel was not really for the server but more for the client intended, so TextChannel could be removed and UDPSocket together with a queue (std::list<string>) could be used. All Sockets can set their destination after construction but before opening.
2) TextChannel (template) can be extended, to take no NodeId. All sockets do support this, and have an added RemoteNode() method to set it later, if it is not already open.

pawel-stanek-ml: Sure. With changes in other places it'll be possible.

SysLogRedirect/SysLogRedirect.h
```diff
@@ -28,48 +28,43 @@ namespace Plugin {
             int wd;
             INotification* notification;
             long int entry_position;
+            string path;
```
pawel-stanek-ml: Note: naming convention is either with leading _  or - for publicly available members - starting with a capital letter.
Adding Init()/Clear() here would also be nice (as opposed to setting/clearing all fields manually). Also CTRO to set them to some decent initial values is also required

SysLogRedirect/SysLogRedirect.h
```diff
@@ -78,72 +73,38 @@ namespace Plugin {
                 file.seekg(0, file.end);
                 pos = file.tellg();
             }
-            file.close();
 
             return pos;
         }
 
-        void Register(const string& entry, INotification* notification, const uint32_t events)
+        void LineEmmiter(const string &path, long int position, INotification &notification)
```
pawel-stanek-ml: & next to type and not variable

SysLogRedirect/SysLogRedirect.h
```diff
@@ -195,18 +141,51 @@ namespace Plugin {
...
+                }
+                _adminLock.Unlock();
+            }
+        }
+
+        void Unregister()
+        {
+            _adminLock.Lock();
+
+            inotify_rm_watch(_iNotifyFd, _observer.wd);
```
pawel-stanek-ml: You don't check IsValid() here and have no CTOR in observer so this is a potential problem here because _observer.wd has god knows what value and _iNotifyFd might be invalid too so you leave it all up to how good  inotify_rm_watch copes when passed invalid values in.

SysLogRedirect/SysLogRedirect.cpp
pawel-stanek-ml: I took a look at Datagram socket and NodeId. It indeed seems default NodeId ctro could be used when creating a socket and RemoteNode() (+LocalNode()) could b callled at later point. In such case no ptr is needed but simple object

SysLogRedirect/SysLogRedirect.h
```diff
@@ -201,39 +201,115 @@ namespace Plugin {
...
-
-            TextChannel() = delete;
-            TextChannel(const TextChannel& copy) = delete;
-            TextChannel& operator=(const TextChannel&);
-
+        class TextChannel : public Core::SocketDatagram {
+            static constexpr uint16_t MAX_BUFFER_LENGHT = 1024;
+            static constexpr uint16_t TIMEOUT_MS = 0;
         public:
             TextChannel(const Core::NodeId& remoteNode)
```
pawel-stanek-ml: This could be default with no params and SocketDatagtam ctro be called with Core::NodeId().Origin() and Core::NodeId().

SysLogRedirect/SysLogRedirect.h
```diff
@@ -201,39 +201,115 @@ namespace Plugin {
...
 
-            void Send(const string& text) override {}
+            // Signal a state change, Opened, Closed or Accepted
+            virtual void StateChange() {}
 
-            void StateChange() override {}
+            void SetDestination(const string& binding, const uint16_t &port) {
+
+                Core::NodeId logNode(binding.c_str(), port);
+                RemoteNode(logNode);
```
pawel-stanek-ml: LocalNode(nodeID.Origin())?

SysLogRedirect/SysLogRedirect.cpp
pwielders: yip

SysLogRedirect/SysLogRedirect.h
pawel-stanek-ml: If not making it as suggested make it explicit at least.

SysLogRedirect/SysLogRedirect.h
```diff
@@ -0,0 +1,360 @@
+#pragma once
+
+#include "Module.h"
+#include <core/core.h>
+
+#include <sys/inotify.h>
+#include <errno.h>
+#include <iostream>
```
pawel-stanek-ml: needed?

SysLogRedirect/SysLogRedirect.h
```diff
@@ -25,51 +25,55 @@ namespace Plugin {
...
 
+    public:
         FileSystemObserver()
             : _adminLock()
             , _iNotifyFd(inotify_init1(IN_NONBLOCK))
-            , _observedEntries()
+            , _observer()
             , _fullFileMode(false)
         {
+            _observer.Reset();
```
pawel-stanek-ml: Note: I'd still add CTOR and call Reset() there instead

SysLogRedirect/SysLogRedirect.h
```diff
@@ -157,25 +130,10 @@ namespace Plugin {
...
-                                    while (std::getline(file, str)) {
-                                        if(str.size() > 0) {
-                                            index->second.notification->Updated(str);
-                                        }
-                                    }
-                                }
-                                file.close();
-                                index->second.entry_position = GetCurrentPossition(index->first);
+                        if (_observer._wd == wd) {
+                            LineEmmiter(_observer._path, _observer._entryPosition, _observer._notification);
```
pawel-stanek-ml: NOTE: Name suggests it'a an object no action

SysLogRedirect/SysLogRedirect.h
```diff
@@ -25,51 +25,55 @@ namespace Plugin {
...
             return (_iNotifyFd != -1);
         }
 
-        inline void Initialize(bool fullFile)
+        void Initialize(bool fullFile, const string& entry, INotification* notification, const uint32_t events)
         {
             Core::ResourceMonitor::Instance().Register(*this);
             _fullFileMode = fullFile;
+
+            Register(entry, notification, events);
```
pawel-stanek-ml: This still looks redundant. I'd put all to Register() and call that

SysLogRedirect/SysLogRedirect.h
```diff
@@ -25,51 +25,55 @@ namespace Plugin {
...
         }
 
-        inline void Uninitialize(void)
+        void Uninitialize()
         {
+            Unregister();
+
             Core::ResourceMonitor::Instance().Unregister(*this);
             _fullFileMode = false;
         }
```
pawel-stanek-ml: Same as above

SysLogRedirect/SysLogRedirect.h
```diff
@@ -25,51 +25,55 @@ namespace Plugin {
...
         {
+            _observer.Reset();
         }
 
-    public:
-        static FileSystemObserver& Instance()
-        {
-            static FileSystemObserver singleton;
-            return (singleton);
-        }
```
pawel-stanek-ml: This not being a signletor is good - otherwise it was bogusly sharing stuff. But.
I think now we have a situation every FileObserver has sepearate inotify having one watch.
I wonder if we could have one inotify registered as IResource and this having multiple watches. @Pierre?

SysLogRedirect/SysLogRedirect.h
Damian-Danylko: Done

SysLogRedirect/SysLogRedirect.h
Damian-Danylko: Fixed

SysLogRedirect/SysLogRedirect.h
Damian-Danylko: Done

SysLogRedirect/SysLogRedirect.h
Damian-Danylko: Done

SysLogRedirect/SysLogRedirect.h
Damian-Danylko: Done

SysLogRedirect/SysLogRedirect.h
Damian-Danylko: Done

SysLogRedirect/SysLogRedirect.h
pawel-stanek-ml: If I read the documentation right it should be possible.

pawel-stanek-ml:
Apart from this one architectual bit to think about (https://github.com/WebPlatformForEmbedded/ThunderNanoServices/pull/191#discussion_r330951984) the code looks ok to me now.

SysLogRedirect/SysLogRedirect.h
pwielders: @pawel-stanek-ml just updated Damian, I think we need the singleton ;-)

pawel-stanek-ml: @pwielders In some form, yes but definitely not in the form it was there before.

FileTransfer/FileTransfer.h
```diff
@@ -0,0 +1,523 @@
...
+                    ::close(_notifyFd);
+                }
+            }
+
+        public:
+            bool IsValid() const
+            {
+                return (_notifyFd != -1);
+            }
+            bool Register(ICallback *callback, const string &filename, const uint32_t type)
```
pwielders: Remove the type as the type and hardcode, where the type is used: IN_CLOSE_WRITE as differnt methods for different files are not supported now. Don't even know what this paremeter does ;-)

Damian-Danylko: Done

PR #199, opened 2019-10-11
[NetworkControll ] Fixed using Open() instead of Create() while saving leases

mipo57:
SaveLeases() function was mistakenly using Open() instead of Create() - that caused the lease file not to be truncated before saving. 

This could lead to a corrupted file if a new version of the file was shorter than the previous one.
Also changed "leaseFile" to "_leaseFile" to better follow convention.

Caught by @sebaszm


mipo57:
The idea was that this way calling constructor of File() every time on saving/loading lease could be avoided for a little memory overhead. But I agree, it might be a little confusing and is called rarely - It is now changed :)

PR #201, opened 2019-10-16
FILETRANSFERCLIENT] File Transfer Linux client

Damian-Danylko:


examples/FileTransferClient/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+find_package(jsonrpc REQUIRED)
```
pwielders: I do not think this dependency is needed. The only dependency you are using is Core, whihc is oke, but please remove this dependency.

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -0,0 +1,295 @@
+#ifndef MODULE_NAME
+#define MODULE_NAME FileTransferClient
+#endif
+
+#include <core/core.h>
+#include <sys/inotify.h>
```
pwielders: Don't think you need inotify.h, list is also not not needed.

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+using namespace WPEFramework;
+
+namespace WPEFramework {
+
+    class TextConnector : public Core::SocketDatagram {
+
+        class FileUpdate : public Core::Thread {
+            private:
+                FileUpdate() = delete;
```
pwielders: According to the latest, changed coding rules, new code should have these deleted constructors public.

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -0,0 +1,295 @@
...
+            private:
+                TextConnector& _parent;
+                std::string _path;
+                std::list<std::string> _newLineQueue;
+        };
+
+        private:        
+            static constexpr uint16_t MAX_BUFFER_LENGHT = 50*1024;
+
+            TextConnector() = delete;
```
pwielders: New coing guideline need to be public

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+                _adminLock.Unlock();
+                
+                return receivedSize;
+            }
+
+            uint16_t SendData(uint8_t *dataFrame, const uint16_t maxSendSize) override
+            {
+                // Do not do anything here
+                return maxSendSize;
```
pwielders: Suggest returning 0 here. Now if we would trigger a send it will continously (loop) to send data as you are now stating that you fully filled the buffer. There is nothing to send so the return val;ue should be 0.

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -0,0 +1,295 @@
...
+            }
+
+            void StateChange() override
+            {
+                // Do not do anything here
```
pwielders: Maybe a notification is usefull, if (IsOpen() ) { printf ("Observing file\n"); } else {printf("No longer observing file\n")} or something similar, just so the user is informed.

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -0,0 +1,295 @@
...
+                _update.Stop();
+
+                _receivedQueue.clear();
+                Close(Core::infinite);
+            }
+
+        public: 
+            uint16_t ReceiveData(uint8_t *dataFrame, const uint16_t receivedSize) override
+            {
+                _adminLock.Lock();
```
pwielders: Reduce the locking to only the emplace_back. All others can go outside of the lock

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -0,0 +1,295 @@
...
+                    Core::Thread::Block();
+                    Wait(Core::Thread::INITIALIZED | Core::Thread::BLOCKED | Core::Thread::STOPPED, Core::infinite);
+
+                    _newLineQueue.clear();
+                }
+
+            private:
+                virtual uint32_t Worker()
+                {
+                    if (_parent.WaitForNewLine(Core::infinite) == true) {
```
pwielders: Here you are loosing me. What is the added value of the temporary storage in the queue? Why not write directly to the file. It is already stored in a queue in the text connector. If there is a good reason for it, why is the Queue than not on this worker stack ?
I think here it might be usefull to keep the file open , std::ofstream a member variable in stead of the stack varaiable it is now, now it will constantly open and close the file on each write 
I also think you could minimize the exposure of the textConnector (parent) by only exposing a method "bool GetLine(string& line, const uint32_t waitTime)" as long as the result of this GetLine returns, true, you write the test in the first parameter to the file.

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -0,0 +1,295 @@
...
+                // Do not do anything here
+                return maxSendSize;
+            }
+
+            void StateChange() override
+            {
+                // Do not do anything here
+            }
+
+            bool WaitForNewLine(const uint32_t time)
```
pwielders: As suggested in the worker above, move this to 1 method, called Getline than the whole locking can be hidden for the external interface. A small interface is simpler to understand.

examples/FileTransferClient/CMakeLists.txt
Damian-Danylko: Fixed

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Fixed

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Fixed

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Fixed

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Fixed

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Fixed

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Added

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Fixed

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Fixed

Damian-Danylko:
Ready for final review

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -101,55 +85,65 @@ namespace WPEFramework {
...
+                } else {
+                    printf("No longer observing file\n");
+                }                
             }
 
-            bool WaitForNewLine(const uint32_t time)
+            bool WaitForNewLine(const uint32_t waitTime)
             {
-                uint32_t status = _newLineSignal.Lock(time);
+                uint32_t status = _newLineSignal.Lock(waitTime);
```
pwielders: This is a very scary situation.... What if, after reporting a timeout, this gets interrupted, the status == timed out, the Commmunication thread, sets the event (since it is running in the next slice, and after continuation of this thread, you reset the event......
Maybe its good to use a thread safe queu, found here: https://github.com/WebPlatformForEmbedded/Thunder/blob/master/Source/core/Queue.h
The only thing you need to share with the FileUpdate, would be this queue

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -101,55 +85,65 @@ namespace WPEFramework {
...
             }
-
+        public: 
             void Unlock()
             {
                 _newLineSignal.Unlock();
             }
 
-            void GetLine(std::string& text, uint16_t& size)
+            bool GetLine(std::string& text, const uint32_t waitTime)
```
pwielders: Right now, it is accessed from only 1 thread, but If you start acessing this from multiple threads at the same time, you must make the WaitForNewLine and extracting the NewLine atomic otherwise you do not know if the line you intended to receive is actually taken by you or some other thread who got the go...

examples/FileTransferClient/CMakeLists.txt
```diff
@@ -1,19 +1,23 @@
-find_package(jsonrpc REQUIRED)
+set(PLUGIN_NAME FileTransferClient)
 
-add_executable(FileTransferClient FileTransferClient.cpp)
+find_package(${NAMESPACE}Plugins REQUIRED)
```
pwielders: I still think that the Core is the only dependency you should have. I do not see any other functionality accept for the functionality comming from Core. So I hope and expect that the only dependency you need here is find_package(${NAMESPACE}Core REQUIRED)

Damian-Danylko: Fixed

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Replaced Event by QueueType implementation

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: I do not see a reason to allow access from multi-threads, maybe I do not see something. Can we talk about it.

Damian-Danylko: According to our conversetion I provided a new solution to simplify an interface

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -14,15 +14,19 @@ namespace WPEFramework {
 
     class TextConnector : public Core::SocketDatagram {
 
+        private:
+            typedef std::shared_ptr<string> EventMessagePtr;
```
pwielders: There is no need to wrap this. The string is already reference counted, so just place the string directly in the queue: 
typedef Core::QueueType<string> EventsQueue;

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -34,34 +38,35 @@ namespace WPEFramework {
 
                 void Stop()
```
pwielders: This Stop is masking the Core::Thread::Stop. Suggest you remove the whole Stop Method, just place the following lines in the destructor of FileUpdate:
Thread::Stop();
_lineQueue.Disable();
Wait(Core::Thread::STOPPED, Core::infinite);

And you are done...

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -34,34 +38,35 @@ namespace WPEFramework {
...
+                        bool status = _lineQueue.Extract(event, Core::infinite);
+
                         // Store to the file
-                        _storeFile << line;
-                    } while (lineAvaliability == true);
+                        if (status == true) {
+                            string line = *event;
+                            _storeFile << line;
+                        }
+                    } while (_lineQueue.Length() > 0);
```
pwielders: How about:
virtual uint32_t Worker() {
    string line;
    while (_lineQueue.Extract(line, Core::infinite) == true) { _storeFile << line; }
    return (Core::infinite);
}

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -72,30 +77,22 @@ namespace WPEFramework {
...
-                , _receivedQueue()
-                , _update(*this, path)                 
+                , _update(_newLineQueue, path)                 
             {
                 Open(0);
             }
 
             virtual ~TextConnector()
             {
                 _update.Stop();
```
pwielders: Since the destructor of the FileUpdate does all actions to stop properly no need to call it, just remove this call _update.Stop();

examples/FileTransferClient/FileTransferClient.cpp
```diff
@@ -72,30 +77,22 @@ namespace WPEFramework {
...
                 std::string text(reinterpret_cast<const char*>(dataFrame), static_cast<size_t>(receivedSize));
-                
-                _adminLock.Lock();
-                _receivedQueue.emplace_back(text);   
-                _adminLock.Unlock();                         
-                
-                EventMessagePtr event = std::make_shared<uint16_t>(1);
+                       
+                EventMessagePtr event = std::make_shared<string>(text);
                 _newLineQueue.Post(event);
```
pwielders: Suggest: 
{
    _newLineQueue.Post( string (reinterpret_cast<const char*>(dataFrame), static_cast<size_t>(receivedSize));
    return (receivedSize);
}

Damian-Danylko: Done

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Done

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Done

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Done

examples/FileTransferClient/FileTransferClient.cpp
Damian-Danylko: Done

pwielders:
Looks good now. @0% code reduction compared to the initial code (only looked at the *.cpp) So it must be good now. (y)

PR #206, opened 2019-10-23
Modified to return result value of SetServerCertificate.

neeraj-deshpande:


pawel-stanek-ml:
@pwielders my problem is we do nothing to make the cast valid. I mean: in the open_cdm API we say OpenCDMError functions return is a one of the following:

ERROR_NONE = 0,
ERROR_UNKNOWN = 1,
ERROR_MORE_DATA_AVAILBALE=2,
ERROR_INVALID_ACCESSOR = 0x80000001,
ERROR_KEYSYSTEM_NOT_SUPPORTED = 0x80000002,
ERROR_INVALID_SESSION = 0x80000003,
ERROR_INVALID_DECRYPT_BUFFER = 0x80000004

But from OCDM I can return the following value from OCDM_RESULT enum: OCDM_INVALID_ARG = 0x80070057. This is out of OpenCDMError scope yet open_cdm API will happily cast it and return to the caller.

pwielders:
@pawel-stanek-ml than the pull is not oke, if we already now that we are using values on the the DRM layer (OCDM_RESULT, OCDM_INVALID_ARG) and this value is not mapped on the other layer (OpenCDMError). That should be resolved by mapping it on the other layer as well. We control both layers so we should align both layers. Lets issue a Jira ticket to align this.

pawel-stanek-ml:
This is exactly what I asked Neeraj for and the ticket is https://metrological.atlassian.net/browse/WPE-398

PR #208, opened 2019-10-28
[wificontrol] Because of wrong comparison hidden flag wasn't set

SKumarMetro:


SKumarMetro:
Fixes WPE-327 [WifiControl] SSID filled with zeros

WifiControl/Controller.h
```diff
@@ -202,7 +202,8 @@ namespace WPASupplicant {
         private:
             void SetSSID(const string& ssid)
             {
-                _hidden = (ssid[0] == '\0');
+                static constexpr const TCHAR* HIDDEN_SSID = "\\x00";
+                _hidden = (ssid.compare(0, strlen(HIDDEN_SSID), HIDDEN_SSID) == 0);
```
pwielders: Why not: 
static constexpr const TCHAR HIDDEN_SSID[] = "\\x00";
_hidden = (ssid.compare(0, sizeof(HIDDEN_SSID)-1, HIDDEN_SSID) == 0);
This will prevent a "count" of something that is already static

SKumarMetro: Sure.  Thought of moving HIDDEN_SSID to config file, that time strlen is relevant.

SKumarMetro:
Combined all wifi changes into branch 'suresh/wifi-combined'

PR #211, opened 2019-10-29
RemoteControl: Fixes from NOS testing

rjpdasilva:
A few RemoteControl plugin changes drafted after NOS testing with the latest framework.

Commit 1 (64aeb5a94d6a) is for making the key release timeout configurable via cmake.
Commit 2 (55dce874ab12) is for fixing a new deadlock use case when plugin activation is done by autostarting it on boot.

All of the changes are required for correct operation of the RemoteControl plugin under the use cases required by NOS.

### DIFF
--- a/RemoteControl/CMakeLists.txt
+++ b/RemoteControl/CMakeLists.txt
@@ -1,6 +1,8 @@
 set(PLUGIN_NAME RemoteControl)
 set(MODULE_NAME ${NAMESPACE}${PLUGIN_NAME})
 
+set(PLUGIN_REMOTECONTROL_RELEASE_TIMEOUT 30000 CACHE STRING "Remote control release timeout")
+
 option(PLUGIN_REMOTECONTROL_RFCE "Enable RF4CE functionality." ON)
 
 set(PLUGIN_REMOTECONTROL_RFCE_REMOTE_ID "GPSTB" CACHE STRING "User string, used for greenpeak")
--- a/RemoteControl/RemoteControl.config
+++ b/RemoteControl/RemoteControl.config
@@ -4,7 +4,7 @@ set (preconditions Platform)
 map()
     kv(repeatstart 500)
     kv(repeatinterval 100)
-    kv(releasetimeout 30000)
+    kv(releasetimeout ${PLUGIN_REMOTECONTROL_RELEASE_TIMEOUT})
 end()
 ans(configuration)
 


pwielders:
64aeb5a oke, but trying to analyse: 55dce87 and the analyses does not sound solid. The Block (if you check the code, will not change any state (default BLOCKED, DEACTIVATING insertyed by the Block() call does nothing, not allowed in this state, Write will only leave a single run for nothing (might be an issue to clean the pipeline on the initialize, but it do not think it could casue an issue.
Would be highly interested in the callstack :-) Could not find it on slack. Would like to dive deeper into this.

rjpdasilva:
> The Block (if you check the code, will not change any state (default BLOCKED, DEACTIVATING insertyed by the Block() call does nothing, not allowed in this state, Write will only leave a single run for nothing (might be an issue to clean the pipeline on the initialize, but it do not think it could casue an issue.

The first Block() call doesn't change the state because is not allowed.
But then, from first Pair() call, there's also a Run(), which indeed is allowed and changes state.
If the actual pthread isn't started yet and another Pair() comes (which is what's happening from the plugin's Initialize()) then a second Block() after Run() will be allowed, and change the state to DEACTIVATE.

Then the actual StartThread() is run and will be blocked in the first wait it does:
```
stateObject.WaitState(INITIALIZED | RUNNING | STOPPED | STOPPING, Core::infinite);
```

On the other side, the second LinuxDevice::Block() call will also be blocked on:
```
Wait(Core::Thread::INITIALIZED | Core::Thread::BLOCKED | Core::Thread::STOPPED, Core::infinite);
```

> Would be highly interested in the callstack :-) Could not find it on slack. Would like to dive deeper into this.

I'll provide the callstack.

pwielders:
@rjpdasilva Clear, but than I suggest that this is solved at the core (WaitState in the Thread should also cope with DEACTIVATED properly). There is no need to solve this at a higher level. Agree ?

pwielders:
Oke, got a fixed lined up in the core to solve 55dce87 in the core. Can you split this pull request tio only include 64aeb5a and remove this pull request ?

rjpdasilva:
Agree. Thought of that but didn't want to mess with the core too much in this case.
I'll rework th pr to have only the other commit.

rjpdasilva:
PR reworked to remove the 2nd commit.

PR #212, opened 2019-10-29
IOConnector: Fixes from NOS testing

rjpdasilva:
The current implementation of IExternal::IFactory::Resource(uint32_t id)
in IOConnector plugin checks the "id" with the GPIO::Pin::Identifier()
for all the GPIO pins we have configured and return pointer to GPIO::Pin
object which matches our "id".

But we need to mask the identifier returned by Identifier() method
before use, This is because when we are creating GPIO::Pin objects the
GPIO pin id is masked with a prefix corresponding to the type of
ExternalBase<> template we are creating.

In our case "class Pin" is created from ExternalBase of type
Exchange::IExternal::GPIO and this our gpio pin id will be masked with
IExternal::identification::GPIO which is 0x20000000. See
Interfaces/IExternal.h for more details.

Thus mask the GPIO::Pin::Identifier() before comparison, We have used
the mask as 0xFFFF and not 0x0FFFFFFF following other places in
IOConnector where it is already used.

All of the changes are required for correct operation of the IOConnector plugin under the use cases required by NOS.

### DIFF
--- a/IOConnector/IOConnector.cpp
+++ b/IOConnector/IOConnector.cpp
@@ -161,7 +161,7 @@ namespace Plugin
         Pins::iterator index = _pins.begin();
 
         while ((index != _pins.end()) && (result == nullptr)) {
-            if (index->first->Identifier() == id) {
+            if ((index->first->Identifier() & 0xFFFF) == id) {
                 result = index->first;
                 result->AddRef();
             } else {


pwielders:
As this is a "generic" implementation and the definition of the ID includes future extensions, like pin type, GPIO/ZWave/Zigbee. So my suggestion is to, on the other side, add the GPIO pin enum value to the ID (0x20000000).
The C (COMRPC) interface can be distributed and thus be accessed by a collecting plugin, there the number needs to be unique and according to the IExternal spec.
The reason for the others "not" to have this is because the others can only be accessed "through" the callsign of this plugin.

rjpdasilva:
> As this is a "generic" implementation and the definition of the ID includes future extensions, like pin type, GPIO/ZWave/Zigbee. So my suggestion is to, on the other side, add the GPIO pin enum value to the ID (0x20000000).
> The C (COMRPC) interface can be distributed and thus be accessed by a collecting plugin, there the number needs to be unique and according to the IExternal spec.
> The reason for the others "not" to have this is because the others can only be accessed "through" the callsign of this plugin.

Understood and agree.
I'll try to use Resource() masked with the GPIO ID on the client plugin side and get back with results.

pwielders:
:+1:

rjpdasilva:
Confirmed.
This PR can be closed.

pwielders:
Oke will close it

PR #213, opened 2019-10-29
Power: Fixes from NOS testing

rjpdasilva:
The Worker thread in the Power plugin to process requests from
SetState() is running through a while loop continuously without blocking
itself waiting for any event. This is a busy wait loop causing high CPU
usage.

Add a new Event object to make the worker thread wait for events
signalled by SetState() to avoid the busy wait loop.

All of the changes are required for correct operation of the Power plugin under the use cases required by NOS.


pwielders:
Revieuwed the code, fix is good but this needs a bigger update, the thread should be removed (as small as it gets) and should schedule it on a workerpool thread. 
Than also some synchronisation resources can be dropped as well. Internall ticket:
https://metrological.atlassian.net/browse/WPE-409

PR #222, opened 2019-11-08
webkitbrowser: switch to 1080p as startup resolution

moorthy-bs:
to fit webkit browser in fullscreen as compositor launches in 1080p


moorthy-bs:
If missed to mentioned from the yocto recipe, making default to 1080p will make the webkit to fit into full screen as other environment such as Westeros in WPEFramework is launching in 1080p

pwielders:
Currently the default deplyment in most receipies is 720p. If we change it in the CMakefile.txt all these projects, that have not set this variable will now "automagically" become 1080p. This will lead to confusion. If you want, for your deplyment the compositor and the webkit to be 1080p, just set the PLUGIN_COMPOSITOR_RESOLUTION and PLUGIN_WEBKITBROWSER_RESOLUTION to 1080p and you should be good to go. Just do not change the defaults.

moorthy-bs:
> Currently the default deployment in most recipes is 720p. If we change the defaults in the CMakefile.txt all these projects, that have not set this variable will now "automagically" become 1080p. This will lead to confusion. If you want, for your deployment the compositor and the webkit to be 1080p, just set the PLUGIN_COMPOSITOR_RESOLUTION and PLUGIN_WEBKITBROWSER_RESOLUTION to 1080p in your recipe and you should be good to go. Just do not change the defaults.

okay. Closing this PR.

PR #226, opened 2019-11-13
enable cursor on westeros compositor

moorthy-bs:
- introduced new field in json config for cursor file
    - enabling cursor only if any file mentioned in the json config
    - reading PNG image in ARGB888 format as westeros expects

### DIFF
--- a/Compositor/Compositor.config
+++ b/Compositor/Compositor.config
@@ -119,6 +119,7 @@ map()
 
     endif ()
 
+    kv(cursor ${PLUGIN_COMPOSITOR_CURSOR_FILE})
 end()
 ans(configuration)
 
--- a/Compositor/lib/Wayland/Westeros.cpp
+++ b/Compositor/lib/Wayland/Westeros.cpp
@@ -1,6 +1,8 @@
 #include "Wayland.h"
 #include <westeros-compositor.h>
 
+#include <png.h>
+
 namespace WPEFramework {
 namespace Westeros {
 
@@ -16,6 +18,7 @@ namespace Westeros {
         Compositor(const string& renderModule, const string& display, const uint32_t width, const uint32_t height)
             : _compositor(nullptr)
             , _virtualKeyboardHandle(nullptr)
+            , _cursor(nullptr)
         {
             TRACE(Trace::Information, (_T("Starting Compositor renderModule=%s display=%s"), renderModule.c_str(), display.c_str()));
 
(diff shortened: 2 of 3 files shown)


moorthy-bs:
Mouse event handling yet to upstream. implementation and testing is under progress

Compositor/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ set(VERSION_REVISION 0)
 
 set(VERSION ${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_REVISION})
 
+find_package(PNG REQUIRED)
```
pwielders: As the cursor is only available in WesterOS suggest to move this dendency to the proper place, begin: Compositor/lib/Wayland/CMakeLists.txt and not in this CMakeList.txt as now aslo the RPI and Nexus compositorrequire PNG but they do not have any curso Support

Compositor/Compositor.config
```diff
@@ -119,6 +119,7 @@ map()
 
     endif ()
 
+    kv(cursor ${PLUGIN_COMPOSITOR_CURSOR_FILE})
```
pwielders: I guess thhis needs to be condition, only if ${PLUGIN_COMPOSITOR_CURSOR_FILE} != "" we add this keyword.

Compositor/lib/Wayland/Westeros.cpp
```diff
@@ -161,9 +165,113 @@ namespace Westeros {
         {
         }
 
+        static unsigned char* readPNGFile(const char *filename, unsigned char &in_width, unsigned char &in_height)
```
pwielders: All our method start with a capital, so could we make this ReadPNGFile ?

Compositor/Compositor.config
moorthy-bs: > I guess thhis needs to be condition, only if ${PLUGIN_COMPOSITOR_CURSOR_FILE} != "" we add this keyword.

okay

Compositor/lib/Wayland/Westeros.cpp
moorthy-bs: > All our method start with a capital, so could we make this ReadPNGFile ?

okay

Compositor/CMakeLists.txt
moorthy-bs: > As the cursor is only available in WesterOS suggest to move this dendency to the proper place, begin: Compositor/lib/Wayland/CMakeLists.txt and not in this CMakeList.txt as now aslo the RPI and Nexus compositorrequire PNG but they do not have any curso Support

Thanks for this finding. Let me update it.

pwielders:
And again a new feature from the community! Thank you !

PR #230, opened 2019-11-15
power: Adds an option in power plugin for not to register for KEY_POWER event

johansaji:
In the original implementation of the power plugin, plugin is
registering for getting KEY_POWER notifications from the remote. Based
on this event power plugin is controlling the power state. In some
cases, the power state changes should be managed by an managing
application/ plugin and in this case, the power key will interfere
with the manager app. So there should be a way to tell the power plugin
not to register for KEY_POWER events.
This commit adds a configuration option to do the same. In the default
build, the power plugin will register for the KEY_POWER. To disable this
feature, one need to set `PLUGIN_POWER_REGISTER_KEY_POWER` to `OFF` from
a recipe.


pwielders:
@rjpdasilva agree, let make this not compile time, but run-time (from the config)

PR #231, opened 2019-11-15
Fix for WPE-438 - Modified Linux Device to send pairing events

neeraj-deshpande:
Modified Pair() and Unpair() methods for KeyDevice class to send pairing events.

However at present, Unpair() method of KeyDevice class is not being called from anywhere. Should it be called in the  KeyDevice destructor.


pwielders:
Looks good. Thanks @rjpdasilva now the pairing of the LinuxInputDevice alseo sends out the pairing events...

PR #232, opened 2019-11-18
JSONRPCClient: Moved SmartClient as SmartLinkType in to thunder JSONRPCLink.h

HaseenaSainul:


examples/JSONRPCClient/JSONRPCClient.cpp
```diff
@@ -9,318 +9,35 @@
...
-	}
-	uint32_t SetProperty(const char method[], const Core::JSON::VariantContainer& object, const uint32_t waitTime = Client::DefaultWaitTime)
-	{
-		return (_connection.Set(waitTime, method, object));
-	}
-	uint32_t GetProperty(const char method[], Core::JSON::VariantContainer& object, const uint32_t waitTime = Client::DefaultWaitTime)
-	{
-		return (_connection.Get<Core::JSON::VariantContainer>(waitTime, method, object));
-	}
+    bool IsActivated() {
```
pwielders: I think this one can go ? It will automagically call the base.

examples/JSONRPCClient/JSONRPCClient.cpp
```diff
@@ -9,318 +9,35 @@
...
-	void Activated()
-	{
-		printf("Monitor Plugin Activated\n");
-	}
-	void Deactivated()
-	{
-		printf("Monitor Plugin Deactivated\n");
-	}
+    void StateChange() override{
+        if (JSONRPC::SmartLinkType<INTERFACE>::IsActivated() == true) {
```
pwielders: If you removed the above, this can be without the namespace.

examples/JSONRPCClient/JSONRPCClient.cpp
```diff
@@ -9,318 +9,35 @@
...
+            printf("%s plugin is activated\n", _callsign.c_str());
+        } else {
+            printf("%s plugin is deactivated\n", _callsign.c_str());
+        }
+    }
 
 private:
-	Connection _connection;
+    string _callsign;
```
pwielders: Why should you maintain the callsign ? It is not neede here, is it ?

HaseenaSainul: It is just used to print the name of the plugin's status requested

PR #233, opened 2019-11-18
[TimeSync] TimeSync sometimes not succeeding

MFransen69:


TimeSync/NTPClient.cpp
```diff
@@ -294,7 +315,11 @@ namespace Plugin {
     void NTPClient::Update()
     {
 
-        _adminLock.Lock();
```
pwielders: Thats scarry.... What if during the "notification" period the element is removed from the list? We are iterating over this list in a moment. If it is not locked, the Register and Unregister can simply add/remove elements.

MFransen69: As was added in the comment the (private) Update method is always used in an already protected context, of course it needs to be locked (I removed it here as I saw other examples in our code base where we do that, indeed double locking does not make it mores safe but decreases performance ;) ).

pwielders: Ohhh missed that the callee already had a lock and here it was taken again...

PR #236, opened 2019-11-21
RemoteControl:LinuxDevice: Send Pairing/UnPairing events based on the…

HaseenaSainul:
… device availablity status

### DIFF
--- a/RemoteControl/LinuxDevice.cpp
+++ b/RemoteControl/LinuxDevice.cpp
@@ -564,7 +564,7 @@ namespace Plugin {
             // We are done, start observing again.
             Run();
 
-            return (true);
+            return (_devices.size() != 0);
         }
         bool Unpair(string bindingId)
         {
@@ -576,7 +576,7 @@ namespace Plugin {
             // We are done, start observing again.
             Run();
 
-            return (true);
+            return (_devices.size() == 0);
         }
 
     private:


RemoteControl/LinuxDevice.cpp
```diff
@@ -658,6 +669,8 @@ namespace Plugin {
             std::vector<IDevInputDevice*>::iterator index = find(_inputDevices.begin(), _inputDevices.end(), inputDevice);
             if (index != _inputDevices.end()) {
                 _inputDevices.erase(index);
+                // Keep in revoked list to clear later on
+                _revokedInputDevices.push_back(inputDevice);
```
pwielders: Why can we not just delete the inputDevice here?

HaseenaSainul: This Clear is calling from the https://github.com/WebPlatformForEmbedded/ThunderNanoServices/blob/73e613d5c71ace4e53c08c491e08b0a3025104d0/RemoteControl/RemoteAdministrator.h#L360(During the deactivation of RemoteControl: https://github.com/WebPlatformForEmbedded/ThunderNanoServices/blob/73e613d5c71ace4e53c08c491e08b0a3025104d0/RemoteControl/RemoteControl.cpp#L280) and if do delete from here it will again invoke https://github.com/WebPlatformForEmbedded/ThunderNanoServices/blob/73e613d5c71ace4e53c08c491e08b0a3025104d0/RemoteControl/RemoteAdministrator.h#L280 from https://github.com/WebPlatformForEmbedded/ThunderNanoServices/blob/73e613d5c71ace4e53c08c491e08b0a3025104d0/RemoteControl/LinuxDevice.cpp#L51

HaseenaSainul: Implementation updated to Release the InputDevices while revoking and not revoke it from the destructor

PR #237, opened 2019-11-21
compositor: mouse event handling

moorthy-bs:
- mouse move and mouse click events handled


moorthy-bs:
is virtualinput_open() give a common handle for different inputs? If so, need to check whether these work on same connector name / nodeId

pwielders:
In the meantime, we had a request to also work with touch and mouse. So on the Virtualinput, we added a way to communicate mouse changes and touch changes.
They all have there own open.

pwielders:
Looks good now. Thanks, will merge!

PR #239, opened 2019-11-24
power: Adds configuration option to register power key and control clients

johansaji:
This commit is for adding an config option to decide on whether to
register for power key or to control the client with IStateControl
interface.


Power/Power.cpp
```diff
@@ -23,11 +23,19 @@ namespace Plugin {
         _power = Core::ServiceAdministrator::Instance().Instantiate<Exchange::IPower>(Core::Library(), _T("PowerImplementation"), static_cast<uint32_t>(~0));
 
         if (_power != nullptr) {
+            Config config;
+
+            config.FromString(_service->ConfigLine());
+            _registerPowerKey = config.RegisterPowerKey.Value();
+            _controlClients = config.ControlClients.Value();
+
+            if (_registerPowerKey) {
```
pwielders: Can we make it a bit more advanced :-) Lets make the RegisterPowerKey a PowerKey (uint16). Than check for IsSet on the PowerKey, If it is Set, the value of this key is also the value you should observe. This allows cusomers to move to any key that could be on the remote to act as a PowerKey. I think now it is hard coded.

Power/Power.cpp
```diff
@@ -55,11 +63,13 @@ namespace Plugin {
...
         _clients.clear();
 
-        // Also we are nolonger interested in the powerkey events, we have been requested to shut down our services!
-        PluginHost::VirtualInput* keyHandler(PluginHost::InputHandler::Handler());
+        if (_registerPowerKey) {
+            // Also we are nolonger interested in the powerkey events, we have been requested to shut down our services!
+            PluginHost::VirtualInput* keyHandler(PluginHost::InputHandler::Handler());
 
-        ASSERT(keyHandler != nullptr);
-        keyHandler->Unregister(&_sink, KEY_POWER);
```
pwielders: This KEY_POWER ==> PowerKey (uint32_t value ? or was it a uint16_t?)

Power/Power.cpp
johansaji: Yaa.. That is a good suggestion... will update  like that

Power/Power.cpp
johansaji: its uint32_t

Power/Power.cpp
johansaji: @pwielders , After giving a thought on it, How do you want to do it. We could only specify the key as an integer in the config, wouldn't it create a confusion?

pwielders: if you do not set a key in the config, the flag IsSet() on the element is false, so that means do not respond on a key. If it has a value set, it is the key value.

rjpdasilva: Hi @pwielders, 

I think @johansaji's question was about having a flat number for the key in the JSON config instead of the key define. For example, for the power key that would be `116` for `KEY_POWER` (see https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h#L191). By seeing just `116` in the config, one may not associated this directly to the actual key.

In my opinion this is not a big issue. One could add some info about this in the cmake variable's help script (it's a linux key code, etc.).

pwielders: Oops missed that one. Indeed lets just add a bit of explanation around this. All the keys that we run through the system are according to the linux input-event-codes. I guess a link to this header file in the CMake should be sufficient.

johansaji: @pwielders , created https://github.com/WebPlatformForEmbedded/ThunderNanoServices/pull/289

PR #243, opened 2019-11-29
Plugin that monitors process associated with any deactivated OOP plugin and kills the process if it does not exit with-in predefined time slot

neeraj-deshpande:


ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+        virtual ~LocalNotification()
+        {
+        }
+
+    public:
+        inline void Open(PluginHost::IShell *service, const uint32_t exittimeout)
+        {
+            ASSERT((service != nullptr) && (_service == nullptr));
+
+            _service = service;
```
MFransen69: Why keep service as a member? It is not used right?
But here is makes sense to have it localy: I would move the register methods to here (as you make it only one notification class), the you control that it is done (responsibility of this class) and you control the order. And then it makes sense to keep it, as you need it for the unregister

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+    public:
+        LocalNotification(ProcessMonitor *parent)
+            : _adminLock()
+            , _job(Core::ProxyType<Job>::Create(this))
+            , _service(nullptr)
+            , _parent(*parent)
+        {
+        }
+        virtual ~LocalNotification()
+        {
```
MFransen69: Perhaps have an assert here in case the Close was not called?

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+            _service = service;
+            _service->AddRef();
+
+            _adminLock.Lock();
+
+            _exittimeout = exittimeout * 1000 * 1000; // microseconds
+
+            _adminLock.Unlock();
+
+            PluginHost::WorkerPool::Instance().Submit(_job);
```
MFransen69: Perhaps better to only submit the job once at least one entry is in the map? Otherwise waist of resources

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+        }
+        inline void Close()
+        {
+            ASSERT(_service != nullptr);
+
```
MFransen69: see comment above, I would add the unregister in this method

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+            PluginHost::WorkerPool::Instance().Schedule(nextSlot, _job);
+        }
+
+        BEGIN_INTERFACE_MAP(LocalNotification)
+        INTERFACE_ENTRY(PluginHost::IPlugin::INotification)
+        END_INTERFACE_MAP
+
+    private:
+        Core::CriticalSection _adminLock;
+        std::map<string, ProcessObject> _processMap;
```
MFransen69: better to use a std::unordered_map

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+        Core::ProxyType<Core::IDispatchType<void>> _job;
+        PluginHost::IShell *_service;
```
MFransen69: Please move type specifiers to the type not the name (improves readability)

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+            else if (currentState == PluginHost::IShell::DEACTIVATION) {
+
+                _adminLock.Lock();
+
+                std::map<string, ProcessObject>::iterator itr(
+                        _processMap.find(service->Callsign()));
+                if (itr != _processMap.end()) {
+                    const uint64_t exitTime = Core::Time::Now().Ticks()
+                            + _exittimeout;
+                    itr->second.Deactivate(exitTime);
```
MFransen69: only submit the job here when there is at least one entry, and schedule it for the correct timeout

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+                            ProcessObject(processId)));
+
+            _adminLock.Unlock();
+        }
+
+        void RemoveProcess(const string callsign)
+        {
+            _adminLock.Lock();
+
+            _processMap.erase(callsign);
```
MFransen69: Guess this might make it not reliable: if it reports deactivated the process still could be running right? We should not remove it here but only remove it once it has been determined in the job the process is not running anymore, correct?

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+            _adminLock.Unlock();
+        }
+
+        void EvalueProcess()
+        {
+            uint64_t currTime(Core::Time::Now().Ticks());
+
+            _adminLock.Lock();
+
+            for (auto itr = _processMap.begin(); itr != _processMap.end();
```
MFransen69: you could consider using for (:) (search for Range-based for loop if it doesn't ring a bell)

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+            }
+            uint64_t ExitTime()
+            {
+                return _exitTime;
+            }
+
+        private:
+            const uint32_t _processId;
+            uint64_t _exitTime;
+            bool _deactivated;
```
MFransen69: I would remove the bool and use _exitTime = 0 to indicate it is not yet deactivated, saves resources

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+            uint64_t currTime(Core::Time::Now().Ticks());
+
+            _adminLock.Lock();
+
+            for (auto itr = _processMap.begin(); itr != _processMap.end();
+                    ++itr) {
+                if (itr->second.Deactivated()) {
+                    if (itr->second.ExitTime() < currTime) {
+                        kill(itr->second.ProcessId(), SIGKILL);
+                    }
```
MFransen69: And instead of doing it on the deactive remove it from the _processmap now

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+                }
+            }
+
+            _adminLock.Unlock();
+
+            uint64_t nextSlot = currTime + static_cast<uint64_t>(1000000);
+            PluginHost::WorkerPool::Instance().Schedule(nextSlot, _job);
```
MFransen69: Only schedule it when there is at least one entry in the list that awaits checking/killing

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+            }
+            else if (currentState == PluginHost::IShell::DEACTIVATED) {
+            }
+        }
+
+        void AddProcess(const string callsign, const uint32_t processId)
+        {
+            _adminLock.Lock();
+
+            _processMap.insert(
```
MFransen69: consider using emplace instead of insert

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+            }
+
+        private:
+            LocalNotification &_parent;
+        };
+
+        class ProcessObject
+        {
+        public:
+            ProcessObject() = delete;
```
MFransen69: I guess you already found out you need the copy construtor for the map. In case you add that also add the assignment operator unless there is a good reason not do that. And as you chose the default implementation indicate that using = default. You could consider to make it movable (would be a very small performance improvement ;) )

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,317 @@
...
+        {
+            uint64_t currTime(Core::Time::Now().Ticks());
+
+            _adminLock.Lock();
+
+            for (auto itr = _processMap.begin(); itr != _processMap.end();
+                    ++itr) {
+                if (itr->second.Deactivated()) {
+                    if (itr->second.ExitTime() < currTime) {
+                        kill(itr->second.ProcessId(), SIGKILL);
```
MFransen69: Although indeed for killing it would not be needed to check up front if the proxess is running I really would like to have the functionality to have a syslog message printed in case a process was still running and needed to be killed so we are aware of the problem and have the root cause fixed

neeraj-deshpande:
Updated as per review comments. 
Created one notifcation class that is derived from both PluginHost::IPlugin::INotification and RPC::IRemoteConnection::INotification.
Removed PluginHost::IWeb and PluginHost::JSONRPC.
Moved register and unregister methods inside _notification.Open() and _notification.Close() respectively.
Modified to submit the job only when necessary and with correct timeout.
Used std::unordered_map instead of std::map.
Modified to removed entry from the std::unordered_map immediately after killing the process.
Deleted copy constructor and assignment operator using =delete where ever applicable.
Used override instead of virtual where ever applicable (including virtual destructor)
Removed unnecessary class member variables.
Added ASSERTS at suggested places.
Modified to use emplace instead of insert.
Added syslog when the process was running and killed.

ProcessMonitor/ProcessMonitor.h
```diff
@@ -183,10 +143,14 @@ class ProcessMonitor: public PluginHost::IPlugin,
...
+                    static_cast<RPC::IRemoteConnection::INotification*>(this));
+
             _service->Release();
             _service = nullptr;
         }
-        virtual void StateChange(PluginHost::IShell *service)
+        void StateChange(PluginHost::IShell *service) override
         {
             PluginHost::IShell::state currentState(service->State());
             if (currentState == PluginHost::IShell::ACTIVATED) {
```
MFransen69: what are the empty if's for? (also see the if (currentState == PluginHost::IShell::DEACTIVATED) below) 
I think they can be removed, agree?

ProcessMonitor/ProcessMonitor.h
```diff
@@ -195,121 +159,116 @@ class ProcessMonitor: public PluginHost::IPlugin,
...
+            if (scheduleTime) {
+                PluginHost::WorkerPool::Instance().Schedule(scheduleTime, _job);
+            }
+        }
+        void Activated(RPC::IRemoteConnection *connection) override
+        {
+            RPC::IProcess *proc =
+                    reinterpret_cast<RPC::IProcess*>(connection->QueryInterface(
+                            RPC::IProcess::ID));
+            if (proc) {
```
MFransen69: we have a rule to always compare with a value to improve readability, so please change this to if(proc != nullptr)

ProcessMonitor/ProcessMonitor.h
```diff
@@ -195,121 +159,116 @@ class ProcessMonitor: public PluginHost::IPlugin,
...
-            uint64_t nextSlot = currTime + static_cast<uint64_t>(1000000);
-            PluginHost::WorkerPool::Instance().Schedule(nextSlot, _job);
+            if (scheduleTime) {
+                PluginHost::WorkerPool::Instance().Schedule(scheduleTime, _job);
+            }
+        }
+        void Activated(RPC::IRemoteConnection *connection) override
+        {
+            RPC::IProcess *proc =
+                    reinterpret_cast<RPC::IProcess*>(connection->QueryInterface(
```
MFransen69: There is a convenience template QueryInterface that you can use for this that is easier to read (and a lot less typing ;) ). For example _browser->QueryInterface<PluginHost::IStateControl>()

ProcessMonitor/ProcessMonitor.h
```diff
@@ -195,121 +159,116 @@ class ProcessMonitor: public PluginHost::IPlugin,
...
                 if (itr != _processMap.end()) {
                     const uint64_t exitTime = Core::Time::Now().Ticks()
                             + _exittimeout;
-                    itr->second.Deactivate(exitTime);
+                    itr->second.SetExitTime(exitTime);
                 }
 
                 _adminLock.Unlock();
+
+                PluginHost::WorkerPool::Instance().Submit(_job);
```
MFransen69: Think we only have to do this when the callsign was actually found in te processmap, now we do this always
And I think we should take into account that the job could already be submitted and perhaps a Schedule would be a better solution to only have it triggered at the correct time

ProcessMonitor/ProcessMonitor.h
```diff
@@ -195,121 +159,116 @@ class ProcessMonitor: public PluginHost::IPlugin,
...
-                    if (itr->second.ExitTime() < currTime) {
-                        kill(itr->second.ProcessId(), SIGKILL);
+            auto itr = _processMap.begin();
+            while (itr != _processMap.end()) {
+
+                exitTime = itr->second.ExitTime();
+                if (exitTime) {
+                    if (exitTime < currTime) {
+                        if (kill(itr->second.ProcessId(), SIGKILL) == 0) {
+                            syslog(LOG_NOTICE, "ProcessMonitor killed %s\n",
```
MFransen69: We have a SYSLOG macro we use for this so it will work on all platforms

ProcessMonitor/ProcessMonitor.h
```diff
@@ -195,121 +159,116 @@ class ProcessMonitor: public PluginHost::IPlugin,
...
+                                    itr->first.c_str());
+                        }
+                        itr = _processMap.erase(itr);
+                        continue;
+                    } else {
+                        if (scheduleTime) {
+                            if (scheduleTime > exitTime)
```
MFransen69: please always use { for if's to improve readbility (we do that always in our code)

ProcessMonitor/ProcessMonitor.h
```diff
@@ -195,121 +159,116 @@ class ProcessMonitor: public PluginHost::IPlugin,
...
-            for (auto itr = _processMap.begin(); itr != _processMap.end();
-                    ++itr) {
-                if (itr->second.Deactivated()) {
-                    if (itr->second.ExitTime() < currTime) {
-                        kill(itr->second.ProcessId(), SIGKILL);
+            auto itr = _processMap.begin();
+            while (itr != _processMap.end()) {
+
+                exitTime = itr->second.ExitTime();
+                if (exitTime) {
```
MFransen69: as mentioned always use a comparison in an if (we do this even if it is a bool, so compare with == true or == false)
so in this case change it to if(exitTime != 0)
Please correct this for the whole file

ProcessMonitor/ProcessMonitor.h
MFransen69: I think the algorithm is not completely correct here (an indication might be that the code for the if and else are exactly the same ;) ). I think you want to reschulde fot the earliest exit time. Please check the code here and test it it is correct

ProcessMonitor/ProcessMonitor.h
```diff
@@ -220,6 +221,27 @@ class ProcessMonitor: public PluginHost::IPlugin
                 PluginHost::WorkerPool::Instance().Schedule(scheduleTime, _job);
             }
         }
+        bool ProcessExists(uint32_t pid)
```
MFransen69: AS already discussed by skype please use the functionality provided in Core. OS abstraction is one there, this makes sure all code will work on other platforms as well (this plugin will now not work on Windows). I guess functionality if a process is running is already present (and will probably not use execute a shell command and use awk as that might not even be available on the device)

neeraj-deshpande: OK. Thanks. I will do the changes as per review comments.

neeraj-deshpande:
Modified to submit the job in the beginning and schedule it when the callsign (with non-zero exittime) is found in the processmap.
Added ScheduleJob() method to schedule the job at lowest exittime in the processmap.
Removed custom ProcessExists() method and used Core IsActive() method to check if the OOP process is alive.
Moved * or & next to type (instead of name) to improve readability.
Removed virtual keyword for ProcessObject destructor.
Removed locking in Notification class close() method.
Removed empty "if statements" in Notification class StateChange() method.
Modified "if statements" to always compare variable with a value to improve readability (instead of if (var) {})
Added template QueryInterface() to query IProcess interface.
Used portable macro SYSLOG (instead of syslog).
Added {} for all "if statements" to improve readability.

ProcessMonitor/ProcessMonitor.h
MFransen69: Could you please also have a look at this comment?

ProcessMonitor/ProcessMonitor.h
MFransen69: Is the submit of the job needed here?

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,284 @@
...
+                    if ((scheduleTime == 0) || (scheduleTime > exitTime)) {
+                        scheduleTime = exitTime;
+                    }
+                }
+            }
+
+            _adminLock.Unlock();
+
+            if (scheduleTime != 0) {
+                PluginHost::WorkerPool::Instance().Schedule(scheduleTime, _job);
```
MFransen69: I must admit I don't know by heart but do we need a revoke for the job first? What would happen if the job was scheduled already (can happen when ScheduleJob() was called from StateChange)? Please check (as mentioned I just don't know could be that you can just reschedule without revoking first)
In case we would need a revoke think about the locking as then ScheduleJob() could be called from mutliple threads which could do a revoke and Schedule in parallel of course if not locked

neeraj-deshpande:
Updated as per review comments.
Called Revoke() before scheduling new job and added Revoke() & Schedule() in the lock.
Marked Copy Constructor and  Assignment Operator as defaulted in ProcessObject class using  ‘=default;’ specifier.

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,284 @@
...
+            _service->Register(
+                    static_cast<RPC::IRemoteConnection::INotification*>(this));
+
+            _exittimeout = exittimeout * 1000 * 1000; // microseconds
+        }
+        inline void Close()
+        {
+            ASSERT(_service != nullptr);
+
+            PluginHost::WorkerPool::Instance().Revoke(_job);
```
MFransen69: Sorry missed that last time but the order here is incorrect. If you Revoke the job here of course as the unregister done later between here and the Unregister, a new Schedule call because of a Notify could be triggered between here and that could post a job (and then also be in conflict with the unprotected _processMap.clear(); ).

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,284 @@
...
+
+            _adminLock.Unlock();
+
+            ScheduleJob();
+        }
+        void ScheduleJob()
+        {
+            uint64_t scheduleTime = 0;
+
+            _adminLock.Lock();
```
MFransen69: Small optimization: as ScheduleJOb is now completely in the Lock it would be faster just to call ScheduleJob always from within a Lock and remove the  _adminLock.Lock(); and  _adminLock.Unlock(); calls. That will be faster then getting the Lock twice

ProcessMonitor/ProcessMonitor.h
```diff
@@ -0,0 +1,284 @@
...
+
+            _adminLock.Unlock();
+        }
+        void Activated(RPC::IRemoteConnection* connection) override
+        {
+            RPC::IRemoteConnection::IProcess* proc =
+                    connection->QueryInterface<RPC::IRemoteConnection::IProcess>();
+            if (proc != nullptr) {
+                AddProcess(proc->Callsign(), connection->RemoteId());
+            }
```
MFransen69: A Release on the IProcess interface is missing, the Interface returned from QueryInterface needs to be released

neeraj-deshpande:
Modified as per review comments. 
Modified Notification Close() method to change the order of job revoke() & Notification unregister.
Removed lock acquisition in ScheduleJob() and called ScheduleJob() from within the lock.
Modified to release IProcess interface after the use.

PR #247, opened 2019-12-05
[WPE-416][NetworkControl] Removed uneccessary references to _parent._…

mipo57:
removed unnecessary references to _parent, as requested in WPE-416.


pwielders:
I think there was more to be cleaned up. Please see Jira ticket

mipo57:
> I think there was more to be cleaned up. Please see Jira ticket
I forgot about Skype message. Now everything should be done ;)

PR #251, opened 2019-12-10
DisplayInfo: plugin minor correction & DeviceInfo: gpu total/free ram status properties are removed

HaseenaSainul:


DisplayInfo/DevicePlatform.h
```diff
@@ -0,0 +1,18 @@
...
+#include "Module.h"
+
+namespace WPEFramework {
+namespace Plugin {
+
+struct IDevicePlatform {
+
+    virtual ~IDevicePlatform() {}
+
+    virtual uint64_t TotalGpuRam() const;
```
pwielders: Let make this an actual interface. Pure abstract methods ( = 0)

DisplayInfo/DevicePlatform.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "Module.h"
+
+namespace WPEFramework {
+namespace Plugin {
+
+struct IDevicePlatform {
```
pwielders: To avoid confusion with the DeviceIdentification and DeviceInfo, can we call this IGraphicsProperties? To better inidicate its purpose?

DisplayInfo/DevicePlatform.h
HaseenaSainul: Sorry I missed that in my first commit, but corrected it later one

DisplayInfo/DevicePlatform.h
HaseenaSainul: Sure, will change

PR #258, opened 2019-12-17
Corrected MemoryObserver instance creation and enabled it.

neeraj-deshpande:


Spark/Spark.cpp
```diff
@@ -47,9 +47,12 @@ namespace Plugin {
                 _spark = nullptr;
             } else {
 
-                _memory = WPEFramework::Spark::MemoryObserver(_connectionId);
+                const RPC::IRemoteConnection *connection = _service->RemoteConnection(_connectionId);
+                ASSERT(connection != nullptr);
 
+                _memory = WPEFramework::Spark::MemoryObserver(connection->RemoteId());
```
MFransen69: I think it would be better to wrap the usage of the connection inside an if(connection != nullptr) to prevent seg faults in release builds

Spark/Spark.cpp
```diff
@@ -47,9 +47,12 @@ namespace Plugin {
...
                 ASSERT(_memory != nullptr);
+                _memory->Observe(connection->RemoteId());
```
MFransen69: Could it be that you are missing a connectecion->Release() here somewhere?

neeraj-deshpande:
Modified as per review comments. 
Wrapped usage of IRemoteConnection inside the condition “if (connection != nullptr)” to prevent seg fault.
Released the IRemoteConnection after usage.

PR #263, opened 2020-01-03
DisplayInfo Plugin updated to support DeviceFirmwareInfo and ConnectionProperties

HaseenaSainul:


DisplayInfo/DisplayInfo.config
```diff
@@ -1,3 +1,11 @@
 set(autostart true)
 set(preconditions Platform)
 
+map()
+    key(root)
```
pwielders: I guess this key can be removed ?

DisplayInfo/DisplayInfo.cpp
```diff
@@ -74,8 +99,23 @@ namespace Plugin {
...
-        displayInfo.Freegpuram = _device->FreeGpuRam();
+        displayInfo.Firmwareversion = _device->FirmwareVersion();
+        displayInfo.Chipset = _device->Chipset();
+
+        Exchange::IGraphicsProperties* graphics(_device->QueryInterface<Exchange::IGraphicsProperties>());
+        displayInfo.Totalgpuram = graphics->TotalGpuRam();
+        displayInfo.Freegpuram = graphics->FreeGpuRam();
+        graphics->Release();
+
+        Exchange::IConnectionProperties* connection(_device->QueryInterface<Exchange::IConnectionProperties>());
```
pwielders: Why mot use _connectionProperties ?

DisplayInfo/DisplayInfo.cpp
```diff
@@ -74,8 +99,23 @@ namespace Plugin {
 
     void DisplayInfo::Info(JsonData::DisplayInfo::DisplayinfoData& displayInfo) const
     {
-        displayInfo.Totalgpuram = _device->TotalGpuRam();
-        displayInfo.Freegpuram = _device->FreeGpuRam();
+        displayInfo.Firmwareversion = _device->FirmwareVersion();
+        displayInfo.Chipset = _device->Chipset();
+
+        Exchange::IGraphicsProperties* graphics(_device->QueryInterface<Exchange::IGraphicsProperties>());
```
pwielders: To safe a continues QueryInterface, you could get and maintain (like _connectionProperties) an interface pointer in the INitialize and free in the Deinitialize ?

DisplayInfo/Nexus/PlatformImplementation.cpp
```diff
@@ -95,16 +249,137 @@ class NexusPlatform : public Plugin::IGraphicsProperties {
...
+        }
+    }
+    static void Callback(void *cbData, int param)
+    {
+        DisplayInfoImplementation* platform = static_cast<DisplayInfoImplementation*>(cbData);
+
+        switch (param) {
+        case 0:
+        case 1: {
+            platform->Submit();
```
pwielders: Lets do the UpdateDisplayInfo() /* no parmaters as this is a private method here, this method knows the class */ . This means that if someone (not based on the Update callback) cals any of these methods, they do have the latest greatest data.

DisplayInfo/Nexus/PlatformImplementation.cpp
```diff
@@ -1,31 +1,96 @@
...
+        }
+        virtual ~Job()
+        {
+            PluginHost::WorkerPool::Instance().Revoke(Core::ProxyType<Core::IDispatch>(*this));
+        }
+
+    public:
+        void Submit()
+        {
+            PluginHost::WorkerPool::Instance().Submit(Core::ProxyType<Core::IDispatch>(*this));
```
pwielders: Prevent multiple times firing of this Job. If one is already pending, no need to submit another one.

DisplayInfo/RPI/PlatformImplementation.cpp
```diff
@@ -1,26 +1,89 @@
...
+        }
+        virtual ~Job()
+        {
+            PluginHost::WorkerPool::Instance().Revoke(Core::ProxyType<Core::IDispatch>(*this));
+        }
+
+    public:
+        void Submit()
+        {
+            PluginHost::WorkerPool::Instance().Submit(Core::ProxyType<Core::IDispatch>(*this));
```
pwielders: See Nexus Implementation :-)

DisplayInfo/RPI/PlatformImplementation.cpp
```diff
@@ -65,12 +228,12 @@ class RPIPlatform : public Plugin::IGraphicsProperties {
         }
 
         // Create string from buffer.
-        value = std::string (equal);
+        value = string(equal);
```
pwielders: Make this Core::ToString(equal) (See https://github.com/WebPlatformForEmbedded/Thunder/blob/master/Source/core/Serialization.h#L50)

DisplayInfo/RPI/PlatformImplementation.cpp
```diff
@@ -95,16 +258,95 @@ class RPIPlatform : public Plugin::IGraphicsProperties {
...
+        DisplayInfoImplementation* platform = static_cast<DisplayInfoImplementation*>(cbData);
+        ASSERT(platform != nullptr);
+
+        if (platform != nullptr) {
+            switch (reason) {
+            case VC_HDMI_UNPLUGGED:
+            case VC_SDTV_UNPLUGGED:
+            case VC_HDMI_ATTACHED:
+            case VC_SDTV_ATTACHED: {
+                platform->Submit();
```
pwielders: See Nexus variant :-)

DisplayInfo/DisplayInfo.config
HaseenaSainul: Thank you Pierre for the comments :).
This is required for mode as well like outofprocess

DisplayInfo/Nexus/PlatformImplementation.cpp
```diff
@@ -1,31 +1,107 @@
...
+                _pendingJob = true;
+            }
+            _adminLock.Unlock();
+        }
+
+    private:
+        virtual void Dispatch()
+        {
+            _parent.Run();
+            _adminLock.Lock();
```
pwielders: It is a small gap, but what, if after the Run() a stat change occures but the _adminLock is not yet entered ;-) We miss a state change. Clear the _pwndingJob, prior to executing the Run.
Both scenarios work, but potentially here you could also use std:atomic<bool>. Might make it a bit more readable/faster and smaller (not sure, maybe nice to check ;-)

DisplayInfo/Nexus/PlatformImplementation.cpp
```diff
@@ -63,8 +139,97 @@ class NexusPlatform : public Plugin::IGraphicsProperties {
...
+
+    BEGIN_INTERFACE_MAP(DisplayInfoImplementation)
+        INTERFACE_ENTRY(Exchange::IDeviceProperties)
+        INTERFACE_ENTRY(Exchange::IGraphicsProperties)
+        INTERFACE_ENTRY(Exchange::IConnectionProperties)
+    END_INTERFACE_MAP
+
 private:
-    void UpdateTotalGpuRam(uint64_t& totalRam) const
+    inline void UpdateFirmwareVersion(string& firmwareVersion) const
```
pwielders: Is this the GPU ChipSet/Firmware version ? If so it is on the right spot (Display) if it is the Device (SOC version SOC firmware) I think we should move the IDeviceProperties to the DeviceIdentifiction plugin.

DisplayInfo/RPI/PlatformImplementation.cpp
```diff
@@ -1,26 +1,100 @@
...
+                _pendingJob = true;
+            }
+            _adminLock.Unlock();
+        }
+
+    private:
+        virtual void Dispatch()
+        {
+            _parent.Run();
+            _adminLock.Lock();
```
pwielders: See Nexus implementation, small gap here.

DisplayInfo/Nexus/PlatformImplementation.cpp
HaseenaSainul: You mean it to DeviceIdentification or DeviceInfo plugin, Since these are very generic to all devices ?

pwielders:
@MFransen69 Can you review this as well

PR #264, opened 2020-01-03
OpenCDMi: parse JSON init data for keyids type

moorthy-bs:
- 'keyids' type initdata contains kids in JSON format.
This has to be parsed and convert to 16 byte key id.


OpenCDMi/CENCParser.h
```diff
@@ -262,6 +263,12 @@ namespace Plugin {
                     } else if ((offset == 0) && (data[0] == '<') && (data[2] == 'W') && (data[4] == 'R') && (data[6] == 'M')) {
                         ParseXMLBox(data, size);
                         offset = length;
+                    } else if (memcmp(&data[offset], JSONKeyIds, 8 /*sizeof(JSON_KeyIds)*/) == 0) {
+                        /* keyids initdata type */
+                        TRACE_L1("Initdata contains clearkey's key ids");
+
+                        /* TODO: optionally we can parse this JSON formatted init data */
+                        //ParseJSONInitData(reinterpret_cast<const char*>(data), length);
```
pwielders: Seems it is not finished yet :-)

OpenCDMi/CENCParser.h
```diff
@@ -420,6 +427,30 @@ namespace Plugin {
                 }
         }
 
+        void ParseJSONInitData(const char data[], uint16_t length) {
+            systemType system(CLEARKEY);
+
+            /* FIXME: JSON parser is badly needed */
+            std::string strData(data + 8, length - 8 - 1);
```
pwielders: JSON Parser is part of the core library. Looks like you are expecting an array of keyids. code like:

void ParseJSONInitData(const char data[], uint16_t length) {
    systemType system(CLEARKEY);

    class KeyIds : public Core::JSON::Container {
    public:
        KeyIds() : Core::JSON::Container() , Keys() {
            Add(_T("kids"), &Keys);
        }   
        virtual ~KeyIds() {
        }   

    public:
        Core::JSON::ArrayType<Core::JSON::String> Keys;
    } parser;

    parser.FromString(string(data, length));

    Core::JSON::ArrayType<Core::JSON::String>::ConstIterator index(parser.Keys.Elements());

    while (index.Next() == true) {
        uint8_t data[KeyId::Length()];
        if (Core::FromString(index.Current().Value(), data, sizeof(data)) == KeyId::Length()) {
            AddKeyId(KeyId(system, data, sizeof(data)));
        }
    }
}

OpenCDMi/CENCParser.h
moorthy-bs: Yes, as I didn't have a proper json parser, I didn't call ParseJSONInitData().  Also without adding keyid in this place, playback done as keyid is added while OnKeyStatusUpdate()
Anyways thanks for the Perfect JSON Parser :)

pwielders:
Will you update the pull request with the parser?

moorthy-bs:
> Will you update the pull request with the parser?

Yes, I can update it. In case of 'keyids' the keys are in base64 format as below and it is not length of 16 
"oW5AK5BW43HzbTSKpiu3SQ": "hyN9IKGfWKdAwFaE5pm0qg"
There should be a decoder as like as in [OCDM-Clearkey](https://github.com/WebPlatformForEmbedded/OCDM-Clearkey/blob/master/JSONWebKey.cpp#L68) to get the 16 byte length keyid. Do you think this is a right place to define base64Decode()? or is it fine to update while onKeyStatusUpdate()? or any other way?

moorthy-bs:
It's done finally. Found base64 decoder and implemented. Now this fine to proceed. :)

OpenCDMi/CENCParser.h
```diff
@@ -262,6 +263,11 @@ namespace Plugin {
                     } else if ((offset == 0) && (data[0] == '<') && (data[2] == 'W') && (data[4] == 'R') && (data[6] == 'M')) {
                         ParseXMLBox(data, size);
                         offset = length;
+                    } else if (memcmp(&data[offset], JSONKeyIds, 8 /*sizeof(JSON_KeyIds)*/) == 0) {
```
pwielders: Why did you not use the sizeof(JSON_KeyIds) i.s.o. the 8 ? I think the sizeof is more robust..

moorthy-bs: > Why did you not use the sizeof(JSON_KeyIds) i.s.o. the 8 ? I think the sizeof is more robust..

Using sizeof() gives an invalid application error `error: invalid application of 'sizeof' to incomplete type 'const uint8_t []' {aka 'const unsigned char []'}` as JSONKeyIds is a const unsigned char array defined in CENCParser.cpp.

moorthy-bs:
> Just saw your: Core::URL::Base64Decode. This is the Base64Decode that is specific to URL exceptions, the official Base64 characters '+' and the '/' are replaced here. There is also a full spec Base64Encode and decode, to be found here:
> https://github.com/WebPlatformForEmbedded/Thunder/blob/master/Source/core/Serialization.h#L242-L247
> I think, but I leave it up to your judgement, that you need the latter.
> So the Decode you are using now is expecting - for + and _ for /
> 
> So let me know if this is as expectedand I will merge it

This make sense. I understand there could be trouble if the base64 string contains '+' or '/'. I tried with suggestion Core::FromString() and it is working. Thanks for this finding, actually I didn't notice this Serialize base64 functions ;p Let me update it.

PR #266, opened 2020-01-10
Plugin to get Player properties 

HaseenaSainul:


pwielders:
Added @pawel-stanek-ml to look at the gstreamer stuff as well, he is more knowledgeable on gstreamer than I am.

PlayerInfo/GStreamer/PlatformImplementation.cpp
```diff
@@ -141,11 +147,61 @@ class PlayerInfoImplementation : public Exchange::IPlayerProperties {
...
+            {"audio/mpeg, mpegversion=(int){2, 4}", Exchange::IPlayerProperties::IAudioIterator::AudioCodec::AUDIO_AAC},
+            {"audio/x-ac3", Exchange::IPlayerProperties::IAudioIterator::AudioCodec::AUDIO_AC3},
+            {"audio/x-eac3", Exchange::IPlayerProperties::IAudioIterator::AudioCodec::AUDIO_AC3_PLUS},
+            {"audio/x-opus", Exchange::IPlayerProperties::IAudioIterator::AudioCodec::AUDIO_OPUS},
+            {"audio/x-dts", Exchange::IPlayerProperties::IAudioIterator::AudioCodec::AUDIO_DTS},
+            {"audio/x-vorbis", Exchange::IPlayerProperties::IAudioIterator::AudioCodec::AUDIO_VORBIS_OGG},
+            {"audio/x-wav", Exchange::IPlayerProperties::IAudioIterator::AudioCodec::AUDIO_WAV},
+        };
+
+        GList* audioDecoderFactories = gst_element_factory_list_get_elements(GST_ELEMENT_FACTORY_TYPE_PARSER | GST_ELEMENT_FACTORY_TYPE_MEDIA_AUDIO, GST_RANK_MARGINAL);
```
pawel-stanek-ml: This won't always work. It's not that uncommon for the decoder to be the element taking audio/video format caps and outputting raw format (without need for any parsing). This is why in cobalt I first check DECODER(s) and fall back to PARSER(s) if none is found.

PlayerInfo/GStreamer/PlatformImplementation.cpp
```diff
@@ -141,11 +147,61 @@ class PlayerInfoImplementation : public Exchange::IPlayerProperties {
...
+            {"video/x-h263", Exchange::IPlayerProperties::IVideoIterator::VideoCodec::VIDEO_H263},
+            {"video/x-h264, profile=(string)high", Exchange::IPlayerProperties::IVideoIterator::VideoCodec::VIDEO_H264},
+            {"video/x-h265", Exchange::IPlayerProperties::IVideoIterator::VideoCodec::VIDEO_H265},
+            {"video/mpeg, mpegversion=(int){1,2}, systemstream=(boolean)false", Exchange::IPlayerProperties::IVideoIterator::VideoCodec::VIDEO_MPEG},
+            {"video/x-vp8", Exchange::IPlayerProperties::IVideoIterator::VideoCodec::VIDEO_VP8},
+            {"video/x-vp9", Exchange::IPlayerProperties::IVideoIterator::VideoCodec::VIDEO_VP9},
+            {"video/x-vp10", Exchange::IPlayerProperties::IVideoIterator::VideoCodec::VIDEO_VP10}
+        };
+
+        GList* videoDecoderFactories = gst_element_factory_list_get_elements(GST_ELEMENT_FACTORY_TYPE_PARSER | GST_ELEMENT_FACTORY_TYPE_MEDIA_VIDEO, GST_RANK_MARGINAL);
```
pawel-stanek-ml: Same as aove

PlayerInfo/GStreamer/PlatformImplementation.cpp
pawel-stanek-ml: Also note that having a parserdoes not guarantee the pipeline to be properly constructed - but I guess it's not needed to be worried about here.

HaseenaSainul: Thanks @pawel-stanek-ml, I followed the method used in the webkit, will look on the cobalt case and handle it.

HaseenaSainul: @pawel-stanek-ml, I have updated based on your comments. Could you please cross check it.
Below formats are supporting by the platforms:
RPI: Audio : [MPEG1, MPEG2, MPEG3, MPEG4, AAC, OPUS] Video :[MPEG, H263, H264, VP8]
Nexus: Audio: [MPEG1, MPEG2, MPEG3, MPEG4, AAC, AC3, AC3Plus, DTS, OPUS, VorbisOGG] Video: [MPEG, H263, H264, H265, VP9]

pawel-stanek-ml: It resembles me the code I implemented in Cobalt so looks good to me, obviously :)

HaseenaSainul: @pawel-stanek-ml, yes, I have followed the method you have implemented, so it looks like similar :)

pawel-stanek-ml: It might be good to share it since we now have 2 places using it being this plugin & cobalt.
What do you think @pwielders @HaseenaSainul ?

pawel-stanek-ml: Regarding the list: it looks legit to me from what I remember (the only thing looking strange to me is vp8 on RPI. I though it always comes with with vp9 and additional plugins needs to be turned on. I can't check it in runtime atm because I have no setup running)

HaseenaSainul: In RPI, the decoder check itself returning the VP8 caps support as true

pawel-stanek-ml: Yup. I see in the code OMX supports vp8

pwielders:
Levae the Gstreamer stuff up to @pawel-stanek-ml logic around it is good to Go!

pawel-stanek-ml:
GStreamer stuff is based on what Cobalt does so it's good to go for me too.
We might think to share the code between Cobalt and this plugin instead of copying but it's not a blocker and can be done separately.

HaseenaSainul:
Thanks @pawel-stanek-ml and @pwielders for the feedbacks. As you suggested, we can start another task to move the GStreamer common stuff to reusable code.

PR #267, opened 2020-01-10
Remove floats and double from a Nanoservices code

Damian-Danylko:


Spark/SparkImplementation.cpp
```diff
@@ -570,11 +570,11 @@ namespace Plugin {
                 EXITSCENELOCK();
             }
 
-            virtual void onScrollWheel(float dx, float dy) override
+            virtual void onScrollWheel(int32_t dx, int32_t dy) override
```
MFransen69: Mhmm, this one has an override, gues it is comming from pxWindow or pxIViewContainer? Should we change the interface (I would think this would now not even compile but I guess I'm missing something ;) )

Damian-Danylko: You are right thsi is part of pxWindow interface, we should not changed it. Unfortunatl;y I did not enable Spark plugin to see whether it still compile. I will revert a change

MFransen69:
The dsg parser was removed in the mean time so we cannot merge this pull request

PR #269, opened 2020-01-13
[VolumeControl] Nexus and Stub implementation.

pawel-stanek-ml:
Volume Control implementation (Nexus). Controls volume globally. Tested on Arris.
Stub implementation so it all compiles for rest of the platforms.


VolumeControl/VolumeControl.config
```diff
@@ -0,0 +1,8 @@
+set (autostart true)
+map()
+    key(root)
+    map()
+      kv(outofprocess false)
```
pwielders: Use the new keyword, this is legacy. Use mode, I think local ?

VolumeControl/VolumeControl.h
```diff
@@ -0,0 +1,142 @@
...
+
+        private:
+            VolumeControl& _parent;
+        };
+
+        void Deactivated(RPC::IRemoteConnection* connection);
+        void VolumeChanged()
+        {
+            uint8_t vol = 100;
+            _implementation->Volume(vol);
```
pwielders: Why is the changed setting the volume ?

VolumeControl/VolumeControl.h
```diff
@@ -0,0 +1,142 @@
...
+            event_volumechange(vol);
+        }
+
+        void MutedChanged()
+        {
+            bool muted = false;
+            _implementation->Muted(muted);
```
pwielders: Or this one setting the muted to false ?

VolumeControl/VolumeControlImplementation.cpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    uint32_t VolumeControlImplementation::Volume(const uint8_t volume)
+    {
+        return _platform->Volume(volume);
+    }
+
+    uint32_t VolumeControlImplementation::Volume(uint8_t& vol) const
+    {
+        vol = _platform->Volume();
```
pwielders: I guess you want to return an error here if the vol > 100.

VolumeControl/VolumeControlJsonRpc.cpp
```diff
@@ -0,0 +1,104 @@
...
+        return _implementation->Muted(param.Value());
+    }
+
+    // Event: volumechange - Signals change of the volume
+    void VolumeControl::event_volumechange(const uint8_t& volume)
+    {
+        VolumechangeParamsData params;
+        params.Volume = volume;
+
+        Notify(_T("volumechange"), params);
```
pwielders: I think we need to align this external interface with the COOMRPC interface, so it should also prepare for Colume and Muted as an event

VolumeControl/VolumeControlJsonRpc.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    // Event: mutedchange - Signals change of the muted property
+    void VolumeControl::event_mutedchange(const bool& muted)
+    {
+        MutedchangeParamsData params;
+        params.Muted = muted;
+
+        Notify(_T("mutedchange"), params);
```
pwielders: I think we need to align this external interface with the COMRPC interface, so it should also prepare for Volume and Muted as an event

VolumeControl/VolumeControl.config
```diff
@@ -0,0 +1,8 @@
+set (autostart true)
```
pwielders: Since this has to join with Nexus (if it is the Nexus IMplementation) suggest to add the Platform as a precondition

VolumeControl/VolumeControlJsonRpc.cpp
```diff
@@ -0,0 +1,104 @@
...
+    // Note: The VolumeControl class must inherit from PluginHost::JSONRPC
+
+    private:
+        void RegisterAll();
+        void UnregisterAll();
+        uint32_t get_volume(Core::JSON::DecUInt8& response) const;
+        uint32_t set_volume(const Core::JSON::DecUInt8& param);
+        uint32_t get_muted(Core::JSON::Boolean& response) const;
+        uint32_t set_muted(const Core::JSON::Boolean& param);
+        void event_volumechange(const uint8_t& volume);
```
pwielders: as described belo, for automatic generation later on, I suggest event_volume and event_muted

pwielders: Just understood that the event here does send the current status at the moment of notification. Thats oke, but than we want to same on the COMRPC. So please align so later we can generate the JSONRPC events from the COMRPC.

VolumeControl/VolumeControlJsonRpc.cpp
pawel-stanek-ml: Will rename to "volume"

VolumeControl/VolumeControlJsonRpc.cpp
pawel-stanek-ml: Will rename to "muted"

VolumeControl/VolumeControl.config
pawel-stanek-ml: Ok

VolumeControl/VolumeControl.config
pawel-stanek-ml: Oh :). I stopped at time this was the new way (as opposed to "set(outofprocess true)". I wasn't even aware there's newer one :D

pawel-stanek-ml: Ah. Right. The Whether guy changes ;).

VolumeControl/VolumeControlJsonRpc.cpp
pawel-stanek-ml: Will add sending it upon successful calls (unfortunately I didn't finf anything in NEXUS API equivalent to these.

VolumeControl/VolumeControl.config
pwielders: indeed there is now also something like "on-other-host"

VolumeControl/VolumeControlJsonRpc.cpp
pwielders: You lost me here, for this event you also first call the "current" volume, mutes state before sending the event, I assume you can slo do this for the COMRPC ?

pawel-stanek-ml: Currently the event is not sent to anyone. I'll add sending it for both JSONRPC and COMRPC once volume/muted is set by the plugin.
It's simply I was hoping NEXUS API has some system wide callbacks for volume/muted change but it seems it doesn't so I need to simulate ones.

pawel-stanek-ml: Ach. I now reread and came to realize I probably misunderstand you previously.
What you're afer here is to make Mute() and Volume() from IVolumeControl::INotification to be given parameters to be in line with JSONRPC events, right?

pawel-stanek-ml: Done. Sorry for confusion

pwielders:
Looks good! Fast implementation. Well done !

PR #271, opened 2020-01-14
Modified to enable “SetTop” functionality in the Nexus compositor server.

neeraj-deshpande:
Not tested as I don’t have Broadcom platform.


pwielders:
The change in the zOrder has been added to the refactored compositor, so no need to merge this pull request anymore, it is in now!

PR #278, opened 2020-01-27
[DIAL] Move use closed to 2.2 spec

pawel-stanek-ml:


DIALServer/DIALServer.h
```diff
@@ -117,15 +117,23 @@ namespace Plugin {
...
             virtual void Stop(const string& data) = 0;
+            virtual bool SupportsHideAndShow() const = 0;
+            virtual bool IsHidden() const = 0;
+            virtual uint32_t Show(const string& data) = 0;
+            virtual void Hide() = 0;
             virtual string URL() const = 0;
-            virtual string AdditionalData() const = 0;
+            virtual const AdditionalDataType* AdditionalData() const = 0;
+            virtual void AdditionalData(AdditionalDataType&& data) = 0;
             virtual void Running(const bool isRunning) = 0;
```
pwielders: Interesting, I see it was already there, but what is the difference with Start/Stop and the Running(true /* start */ or false /*stop */ )
I think this is due to the JSONRPC API, control at WebWordl, they will use this, but I think we need to simplify this interface to move this to Start/Stop.

pwielders: Just checked the code (bit below) the Start/Stop is the interface towards the DIAL OTT Client, and this is for the "slave mode" to trigger that it is running , or is stopped. So I guess we need to continue support for it.

DIALServer/DIALServer.h
```diff
@@ -117,15 +117,23 @@ namespace Plugin {
         };
 
         struct IApplication {
+            using AdditionalDataType = std::unordered_map<string, string>;
             virtual ~IApplication() {}
 
             // Methods that the DIALServer requires.
             virtual bool IsRunning() const = 0;
             virtual bool HasAllowStop() const = 0;
```
pwielders: We need to change/use 1 naming supportXXXXX v.s. hasXXXXX. Since we already have some methods (also in other parts of the system) I propose to use hasXXX

DIALServer/DIALServer.h
```diff
@@ -239,9 +288,13 @@ namespace Plugin {
             {
                 return ("");
             }
-            virtual string AdditionalData() const
+            void AdditionalData(AdditionalDataType&& data) override
             {
-                return ("");
+                _additionalData = std::move(data);
+            }
+            virtual const AdditionalDataType* AdditionalData() const
```
pwielders: Why would you return a pointer here? It might suggest a Producer/Consumer message (so it suggests that the callee has to do a delete on this, if he is done with it, but I think, loooking at the code, you should not delete it :-)

DIALServer/DIALServer.h
```diff
@@ -281,14 +334,100 @@ namespace Plugin {
...
+        struct Version {
+            Version(uint16_t major, uint16_t minor, uint16_t patch)
+                : Major(major), Minor(minor), Patch(patch) {}
+            Version() : Version(0, 0, 0) {}
+
+            bool IsValid() const { return Major != 0 || Minor != 0 || Patch != 0; }
+
+            bool IsDefault() const { return Major == kDefaultMajor && Minor == kDefaultMinor && Patch == kDefaultPatch; }
+
+            bool operator<(const Version& other) const {
```
pwielders: There is an official code guideline that tells you to only use 1 return in a method. There are a very few methods that have 2 returns but here are really a lot.

DIALServer/DIALServer.h
pawel-stanek-ml: Yes. Exactly. This is used by JS world in passive mode to inform the DIAL server the app is up & running. We could simplify and assume that sending the event to the JS is enough and the app is running but than we risk the state tracking will be wrong if the app isn't in fact launched/fails to launch for some reason. Do we want take that chance?

DIALServer/DIALServer.h
```diff
@@ -281,14 +334,100 @@ namespace Plugin {
...
+
+                  if (other.Minor == Minor) {
+                      return other.Patch > Patch;
+                  }
+              }
+
+              return false;
+            }
+
+            bool operator==(const Version& other) const {
```
pwielders: Is the operator != automatically generated by calling this one ?

DIALServer/DIALServer.h
```diff
@@ -281,14 +334,100 @@ namespace Plugin {
...
+                return other.Major == Major && other.Minor == Minor && other.Patch == Patch;
+            }
+
+            bool operator<=(const Version& other) const {
```
pwielders: Same question here than >= is also automatically generated? If not we are looking at a-symetrical interfaces.

DIALServer/DIALServer.h
```diff
@@ -281,14 +334,100 @@ namespace Plugin {
...
+              Patch = kDefaultPatch;
+            }
+
+            void Clear() { Major = Minor = Patch = 0; }
+
+            uint16_t Major;
+            uint16_t Minor;
+            uint16_t Patch;
+
+            static constexpr uint8_t kDefaultMajor = 1;
```
pwielders: Interesting, the default type is uint8_t (I guess enough) and the actual members are uint16_t guess we need a bit of aligning here.

DIALServer/DIALServer.h
pwielders: Nope, so agree, we need to mainatan the Running method.

DIALServer/DIALServer.h
pawel-stanek-ml: Nothing will be generated and anyone trying to use anu operator except for the ones provided will face compilation error (but I agree that for completness all usefull operators should be provided just trimmed it to use case since it's simple internal helper)

DIALServer/DIALServer.h
pawel-stanek-ml: Fixed

DIALServer/DIALServer.h
pawel-stanek-ml: Done

DIALServer/DIALServer.h
pawel-stanek-ml: Done

DIALServer/DIALServer.h
pawel-stanek-ml: Fixed

DIALServer/DIALServer.h
pwielders: I guess this one is still left?
Suggest to use "bool HasStartStop() const = 0"for SupportStart and Support Stop, to avoid a strange inconsistency on something you can start but not stop or vice versa. and "bool HasHideAndShow() const = 0"

pawel-stanek-ml: Ach, Right! Forgot about it. On it.

PR #280, opened 2020-01-29
[WPE-542] Make Dial to work with Cobalt (and vice versa).

pawel-stanek-ml:
NOTE: THIS IS BASED ON pstanek/dial-2.2/1 BRANCH (and if simply merged will be merged to that branch).


pawel-stanek-ml:
This makes both good'n'old Webkit based YT and new-kid-on-the-block Cobalt can be used. It doesn't metter which is launched from DIAL POV. The client simply has to respect either the var or IBrowser::SetURL()

pawel-stanek-ml:
Fill file a pull requestwith this rebased

PR #282, opened 2020-02-03
Development/wpe 471

pwielders:
Due to the adaptions in the Threading/ThreadPools/WorkerPools there is now "different functionality available for the plugins to use Threads (from the workerPool) adapt the Job code that existed in the plugins to reflect these changes. The old fashion way (code it all) still works but there are now some easo to use helpers for creating Jobs:
Core::ThreadPool::JobType<IMPLEMENTATION>
Core::WorkerPool::JobType<IMPLEMENTATION>
The latter inherits from the first one but does no longer the developepr to find the WorkerPool. Also the Jobs will make sure that the will never be queued more than once.


DisplayInfo/Nexus/PlatformImplementation.cpp
```diff
@@ -212,7 +212,7 @@ class DisplayInfoImplementation : public Exchange::IGraphicsProperties, public E
         rc = NxClient_GetDisplayStatus(&status);
         if (rc == NEXUS_SUCCESS) {
             connected = status.hdmi.status.connected;
-
+/*
```
MFransen69: See the comment I added below. Something is fishy here. Not related to Pierre's changes but mist be fixed anyway before merging

PR #283, opened 2020-02-04
Compositor:Nexus: add NXSERVER find package also in the dependency of…

HaseenaSainul:
… NEXUS_SERVER_EXTERNAL

### DIFF
--- a/Compositor/lib/Nexus/CMakeLists.txt
+++ b/Compositor/lib/Nexus/CMakeLists.txt
@@ -2,7 +2,6 @@ set(TARGET ${PLATFORM_COMPOSITOR})
 
 message("Setting up ${TARGET} for Nexus platfrom")
 
-add_subdirectory (NexusServer)
 
 find_package(${NAMESPACE}Core REQUIRED)
 find_package(${NAMESPACE}Plugins REQUIRED)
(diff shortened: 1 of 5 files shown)


HaseenaSainul:
It using header file nxserverlib.h to get many of the NXClient functionality as well for both EXTERNAL and INTERNAL case, say : NEXUS_PlatformSettings, NxClient_JoinSettings etc. FindNXServer.cmake is searching for either libnxserver.a or nxserver. it setting the status as found if one of them is available. if nxserver binary available case it is exposing only the header path : [NXSERVER.cmake](https://github.com/WebPlatformForEmbedded/ThunderNanoServices/blob/6b5e2bb60a254cc20e11c50d5c70009b7edc1832/cmake/FindNXSERVER.cmake#L32-L44). So this change made to include this header path also into the include directories (not using NXSERVER linkers).

Other issue noticed with the EXTERNAL case is, the macros condition checks are not properly used many of the places which are not applicable to external server mode: for example, [case1](https://github.com/WebPlatformForEmbedded/ThunderNanoServices/blob/6b5e2bb60a254cc20e11c50d5c70009b7edc1832/Compositor/lib/Nexus/NexusServer/NexusServer.h#L132-L133), [case2](https://github.com/WebPlatformForEmbedded/ThunderNanoServices/blob/6b5e2bb60a254cc20e11c50d5c70009b7edc1832/Compositor/lib/Nexus/NexusServer/NexusServer.cpp#L71-L75), [case3](https://github.com/WebPlatformForEmbedded/ThunderNanoServices/blob/6b5e2bb60a254cc20e11c50d5c70009b7edc1832/Compositor/lib/Nexus/NexusServer/NexusServer.cpp#L87-L90)

Compositor/lib/Nexus/CMakeLists.txt
```diff
@@ -26,11 +26,17 @@ set_target_properties(${TARGET} PROPERTIES
 
 if(NEXUS_SERVER_EXTERNAL)
     find_package(NXCLIENT REQUIRED)
+    find_package(NXSERVER REQUIRED)
+
     target_link_libraries(${TARGET}
         PRIVATE
             ${NAMESPACE}nexusserver::${NAMESPACE}nexusserver
```
pwielders: I think, in stead of adding the NXSERVER lib, I think you can remove the target_link_library name ${NAMESPACE}nexusserver::${NAMESPACE}nexusserver. As it is now running externally and not as a composition of the compositor plugin.

HaseenaSainul: Sure @pwielders, as we discussed I will update this :)

pwielders:
@HaseenaSainul very good cleanup action! Seems like, evn though one selected EXTERNAL_SERVER it was still using a lot of functionality. Love this cleanup!

HaseenaSainul:
Yes, I tested with and without EXTERNAL. without external its working fine. With External also Graphics is working fine, but not able to control UI using the remote keys, observed the same with previous version as well(the version before cleanup).

PR #285, opened 2020-02-06
Fixed OCDM plugin where IRemoteConnection was acquired twice but released only once.

neeraj-deshpande:


OpenCDMi/OCDM.cpp
```diff
@@ -105,22 +105,22 @@ namespace Plugin {
         } else {
             _opencdmi->Initialize(_service);
 
-            if ((_connectionId != 0) && (_service->RemoteConnection(_connectionId) == nullptr)){
+            ASSERT(_connectionId != 0);
+            const RPC::IRemoteConnection *connection = _service->RemoteConnection(_connectionId);
+            ASSERT(connection != nullptr);
```
pwielders: How do you want to gurantee this? This assert might be a false positive. The process might have crashed in the mean time and the connection == nullptr.

neeraj-deshpande:
Removed the ASSERT(connection != nullptr).

PR #291, opened 2020-02-12
Streamer: fix to avoid crash during the stop

HaseenaSainul:
During the stop sequence Player is firing the StateChange event through the decoder Event interface. This interface use registered callback to send to the requesters. But this callback is releasing before detaching the player and not setting it to nullptr

### DIFF
--- a/Streamer/Frontend.h
+++ b/Streamer/Frontend.h
@@ -88,6 +88,9 @@ namespace Player {
                 uint32_t Release() const override
                 {
                     if (Core::InterlockedDecrement(_referenceCount) == 0) {
+
+                        ASSERT(_callback == nullptr);
+
                         if (_callback != nullptr) {
                             _callback->Release();
                         }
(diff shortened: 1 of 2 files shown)


HaseenaSainul:
In this case, we are calling parent.Detach from the destructor of Decoder: [here](https://github.com/WebPlatformForEmbedded/ThunderNanoServices/blob/60ed38b461b777c6f8623328718feb77c4751e16/Streamer/Frontend.h#L80), then it will initiate the Detach Process [here](https://github.com/WebPlatformForEmbedded/ThunderNanoServices/blob/60ed38b461b777c6f8623328718feb77c4751e16/Streamer/Frontend.h#L418). This Detach call will fire StreamerEvent and it call the Event call of decoder@ [here](https://github.com/WebPlatformForEmbedded/ThunderNanoServices/blob/60ed38b461b777c6f8623328718feb77c4751e16/Streamer/Frontend.h#L198-L204). But if we set _callback as nullptr before the invocation of detach(), it will sendup with segfault

HaseenaSainul:
@pwielders, update the code to reset the Callback before doing the release of FrontEnd and Decoder from the Streamer context, since the callback registration is happening from the Streamer context during the creation of Stream/Control proxies.

PR #296, opened 2020-02-24
[DIAL][Cobalt][WPE-556] Enabled DIAL to attatch to Cobalt during runtime

mipo57:
!!! important !!!
Needs at lest cobalt 2b05944b27596a50c2d94b02b358fb289980a990 to work

Implemented a possibility to change Cobalt URL during runtime. This allowed for DIAL to kick in when Cobalt is already running.

### DIFF
--- a/DIALServer/DIALServer.h
+++ b/DIALServer/DIALServer.h
@@ -115,19 +115,51 @@ namespace Plugin {
             using AdditionalDataType = std::unordered_map<string, string>;
             virtual ~IApplication() {}
 
-            // Methods that the DIALServer requires.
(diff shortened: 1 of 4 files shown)


DIALServer/DIALServer.h
```diff
@@ -657,14 +716,19 @@ namespace Plugin {
...
             }
-            inline const string URL() const
+            inline string URL() const
             {
                 return (_application->URL());
             }
+            inline void URL(const string& url) 
+            {
+                _application->URL(url);
+            }
```
pawel-stanek-ml: This is browser centric. Payload might be anything.

